PR #16, opened 2025-02-12
ci: add cdash status

vicentebolea:
This will be provided by circle CI so that even if your Github actions jobs allocation is saturated we still get responsive cdash status.


vicentebolea:
I need to merge this in order for this to be visible in circle ci. Merging

Issue #10, opened 2025-02-06
Setup SPACK CI (UO)


vicentebolea:
* This CI will contain Linux and macOS jobs.

PR #18, opened 2025-02-13
ci: add spack ci and Linux builds

vicentebolea:
Fixes: #10 

This adds coverage for the following features/compilers/os:

- Ubuntu 20.04/22.04/24.04
- Almalinux 8
- GCC 9,10,12,13,14
- CLANG 6/11
- Kokkos
- HIP
- CUDA
- Doxygen
- MSAN sanitizer.


.gitlab-ci.yml
```diff
@@ -247,3 +247,4 @@ include:
   - local: '/.gitlab/ci/ubuntu2004.yml'
   - local: '/.gitlab/ci/ubuntu2204.yml'
   - local: '/.gitlab/ci/windows10.yml'
+  - local: '/.gitlab/ci/sync-spack.yml'
```
kmorel, line 250: Why do we have a `.gitlab` directory? The repo is in github now.

vicentebolea: Spack CI runs in a gitlab instance

kmorel: Got it. So a PR's commit is pushed to the spack GitLab instance and the CI is run from there, correct? And through magic that all gets reported back to the GitHub PR.

vicentebolea:
@kmorel this ready to review. Some of the build have some warnings (which I believe that are good things to fix in our code), I propose to fix this in other iterations.

vicentebolea:
I have also commented ubuntu2204_gcc10_cuda since it was failing and I could not figure out how to fix it.

kmorel (approved):
This looks good to me.

It looks like the spack.io builds are not reporting to CDash. I assume that is happening later.

vicentebolea:
> This looks good to me.
> 
> It looks like the spack.io builds are not reporting to CDash. I assume that is happening later.

@kmorel you are right, I will address this later

PR #21, opened 2025-02-17
lfs: move LFS files to viskores-data submodule

### LINKED ISSUE #19: Small Github free bandwith quota for LFS files
We have already hit the small 1GiB limit for the LFS bandwidth (egress/ingress) as shown below:

Image

We have two course of actions:

1. Pay for additional bandwidth (next plan goes with 50GiB/Month)
2. Move our ~80MiB worth of LFS files to our source repo as regular git blobs.

I vote for the option 2, it will also strongly simplify working with viskores repo

vicentebolea:
Fixes #19 

This moves all the LFS to files  to the a new `viskores-data` repository that it is now a submodule of this repository under the path of `data/`.

The only LFS file that I did not move to the submodule is `tutorial/data/kitchen.vtk`, the reason of this is that users expect the tutorial project to be self-contained, referencing files outside of it subtree would defeat the purpose.

This PR completely removes every mention to git lfs.


docs/users-guide/building.rst
```diff
@@ -84,7 +84,7 @@ The following example shows the steps for the case where the |Viskores| source i
    cmake-gui ../viskores-v2.1.0
 
 .. _fig:CMakeGUI:
-.. figure:: images/CMakeGUIBoth.png
+.. figure::  ../../data/users-guide/images/CMakeGUIBoth.png
```
kmorel, line 87: Hmm. It's kind of awkward to reference images here. Maybe eventually we will make a second submodule for these images. But let's think about that and worry about it later.

kmorel (approved):
So, when kitchen.vtk was pulled out of LFS, it sounds like this one file is now part of the repo. I see your logic for it. As long as this is the exception to the rule, it should probably be OK.

I think you are still working on this PR, but it is looking OK to me.

docs/users-guide/building.rst
vicentebolea: @kmorel thanks for the prompt review, I think that this is ready to go. I will merge it.

Issue #11, opened 2025-02-06
Setup Windows CI jobs


vicentebolea:
* This CI will be hosted in Github Actions.

PR #23, opened 2025-02-18
ci: add windows build

vicentebolea:
Fixes: #11 

This PR introduces a windows job that runs in github actions:

- It uses Visual Studio 16 2019 as a cmake generator.
- It uses corresponding MSVC as the C/CXX compilers. 
- This tests a build that is built without GLUT.
- It blacklists the failing test: `UnitTestVTKDataSetReader`, we can come back to this later.
- It uses the gitlab infrastructure codes through the use of github_adaptor.cmake. I found
  this design pattern to be a better alternative than other possible designs since it
  simplifies ourCI by avoiding reimplementation of the same scripts for the two CI vendors. 
  Since most of our CI will run in a gitlab instance, and our legacy CI was using gitlab, it makes sense to have our CI infrastructure code written for gitlab and use adaptors for each of any other CI vendors.


.github/ci/vcvarsall.ps1
```diff
@@ -12,7 +12,7 @@
 
 $erroractionpreference = "stop"
 
-cmd /c "`"$env:VCVARSALL`" $VCVARSPLATFORM -vcvars_ver=$VCVARSVERSION & set" |
+cmd /c """$env:VCVARSALL"" $env:VCVARSPLATFORM -vcvars_ver=$env:VCVARSVERSION && set" |
```
kmorel, line 15: Are you sure this quoting is correct? I don't know the details of how powershell does quotation grouping, but I would expect a sequence of two quotes `""` to be interpreted as an empty string and be replaced with nothing. Removing the two instances of `""`, you are left with a single level of quotations that quotes the whole thing.

```suggestion
cmd /c "$env:VCVARSALL $env:VCVARSPLATFORM -vcvars_ver=$env:VCVARSVERSION && set" |
```

If that is what you meant, write that. Otherwise, the quotes may need to be changed.

kmorel (commented):
Just a minor question about the correct use of quotations. Otherwise, this looks good to me as long as the CI works.

.github/ci/vcvarsall.ps1
vicentebolea: This is one of the powershell oddities. Inside a `""` string you can use a `""` to signify a `"`. https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-7.5

vicentebolea: And we need to quote VCVARSALL since it is a filepath.

kmorel: Oh, OK. Weird.

PR #27, opened 2025-03-04
Add ccache support to Windows CI build

vicentebolea:
This change adds ccache support to the Windows build in GitHub Actions to:
- Improve build speed by caching compiled objects
- Use GitHub Actions cache to persist ccache data between runs
- Add ccache statistics reporting


vicentebolea:
Closing this since getting ninja+msvc (Visual Studio generator does not support ccache) to build viskores seems to be a more difficult problem

PR #30, opened 2025-03-05
spack,ci: correct reported commit to cdash

vicentebolea:


kmorel (approved):
Looks like it is working. I see the Frank results in CDash now.

PR #32, opened 2025-03-08
license: add DCO comment

### LINKED ISSUE #31: Change check copyright test to check license
The contributing guidelines for Viskores states that contributors do not need to assign copyrights to Viskores. This is a change from our guidelines for VTK-m. Because of this, there is a test in Viskores that violates these guidelines.

The `CopyrightStatement` test checks to make sure that there is a statement that assigns copyright. We should have a similar test, but we need to change the statement that is checked. The name of the test should be changed to `LicenseStatement`. The comment that it checks for should point to [LICENSE.txt](https://github.com/Viskores/viskores/blob/main/LICENSE.txt). It should also declare that all contributions agree to the [Developer Certificate of Origin](http://developercertificate.org/).

The copyright statement that is currently at the top of every file can remain. It will just be below the new license statement.

vicentebolea:
fixes #31 

This PR attaches:

```
The contents of this file are covered by the Viskores license. See
LICENSE.txt for details.

By contributing to this file, all contributors agree to the Developer
Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
```

To the beginning of every source file.

- We also renamed CMake/ViskoresCheckCopyright.cmake` to `CMake/ViskoresCheckLicense.cmake`
- We have also modified `CMake/ViskoresCheckLicense.cmake` to check for the above License statement.


kmorel:
I feel like it should start with a statement like "This file is covered by the Viskores license."

You make a good point about making sure the disclaimer covers what we say in contributing and the technical charter. Once we have something, I might try ask someone at the Linux Foundation about it.

vicentebolea:
Please find the modified License comment in the last push

CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+##============================================================================
+##  This file is covered by the Viskores license, see LICENSE.txt for details.
```
kmorel, line 2: This is a nitpick, but how about phrase it like this:

```suggestion
##  The contents of this file are covered by the Viskores license. See
##  LICENSE.txt for details.
```

kmorel:
Looks good.

What is your plan to push this header to the files? You could change them all now, but you'll have to repeat the process when converting the final version of VTK-m.

vicentebolea:
@kmorel, should we prepend the new License notice at every source code? Or do you want to run this by the HPSF?

kmorel (approved):
+2

PR #34, opened 2025-03-12
ci: add job summaries

vicentebolea:
This helps us keeping track of the ccache performance.


kmorel (approved):
+2

I see no problem with this.

PR #37, opened 2025-03-15
ci: change pull policy in gitlab images

vicentebolea:
We need this to leverage changes in the image pull policy in the UO Frank cluster CI.


kmorel:
> I would say lets increase its timeout since there is value in keeping them running in different environments.

It looks improved, but there still seem to be some timeouts. Might need to increase the timeout for some other tests.

kmorel (approved):

PR #38, opened 2025-03-20
ci: delegate cdash status into the cdash-status action

vicentebolea:
I have recently created a github action that setup the cdash status as shown here: https://github.com/marketplace/actions/cdash-status

In this PR remove the adhoc implementation of the cdash status and delegate into this action. This will help me not to duplicate effort across the different project that uses this.


.github/workflows/external.yml
```diff
@@ -0,0 +1,23 @@
...
+permissions: write-all
+
+jobs:
+  generate_statuses:
+    env:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: vicentebolea/cdash-status@v1
+        project: Viskores
```
kmorel, line 23: According to the documentation at https://github.com/marketplace/actions/cdash-status, this should read

```suggestion
      - uses: vicentebolea/cdash-status@v1
        with:
          project: Viskores
```

kmorel (changes requested):
I don't think that this is working correctly. The `CDash` entry in the checks is no longer showing up.

Screenshot 2025-03-21 at 7 31 52 AM

vicentebolea:
@kmorel we do not have the cdash status since I remove the Circle CI infrastructure code. The reason of the new workflow of not working is that this workflow runs in `pull_request_head` type of events. This is an interesting workaround for Github to run `pull_requests` workflows from forks and sharing secrets with it (env variables) while avoiding leaking secrets. These workflows runs with the workflow files as they are in the main branch (as opposed to the pull request).

kmorel:
@vicentebolea I only understood the first sentence of that.

So, where do you go to find the CDash reporting for a PR?

vicentebolea:
> ?

We can only test this kind of github workflow once it is merge in main, this is by design of `pull_request_head` types of workflows.

kmorel:
> We can only test this kind of github workflow once it is merged in main, this is by design of `pull_request_head` types of workflows.

OK. So, you are saying that after this is merged in, we will see the CDash link in the list of checks. If that is the case, then +1.

PR #39, opened 2025-03-21
Bump diy

vicentebolea:
This sync current VTK-m DIY state with Viskores.


kmorel (approved):
Looks good except for those cuda errors on Frank. I assume those are transient and unrelated.

PR #47, opened 2025-03-28
ci: only run on main,release* branches

vicentebolea:
This is needed to avoid running twice the CI when the branch of Pull request exist in the viskores main repo. Observed from the automated dependbot pull requests.


vicentebolea:

PR #48, opened 2025-03-28
git: sync release branch with main in every push

vicentebolea:


kmorel:
To be clear, every time something is pushed to release, the main branch gets updated with the latest commit as well, right?

vicentebolea:
@kmorel that is true, however the changes are not merged to main (-s ours strategy), this is to mimic our existing workflow in VTK-m. In any case, this is still WIP I need to find a way to test this.

kmorel:
It looks like the new file is failing the `LicenseStatement` test: https://open.cdash.org/tests/1968496381

PR #49, opened 2025-03-30
ci: use distributed cache for ccache

vicentebolea:
This is part of an effort with the UO frank cluster administration to improve resource utilization through using a distributed cache for ccache.


vicentebolea:

PR #52, opened 2025-04-10
docs: re-add images

vicentebolea:


kmorel (approved):
All the images that I checked in the generated documentation looked correct.

PR #57, opened 2025-04-10
docs: add doxygen pages to readthedocs

vicentebolea:
We already use breathe for describing viskores symbols in readthedocs. In addition to that with this change we will host the whole doxygen generated pages in readthedocs.


vicentebolea:
@kmorel it seems that adding the whole doxygen into readthedocs looks certainly odd as shown here: https://viskores--57.org.readthedocs.build/en/57/api.html

vicentebolea:
@kmorel merging this as I can only test it in main. I will revert these changes if it breaks anything.

PR #61, opened 2025-04-16
ci: ccache options

vicentebolea:


kmorel:
I don't know enough about ccache, particularly with how the cache is stored and restored between CI runs. In particular, what are the consequences of setting the maxsize? I assume this is to limit the overhead of transferring data before/after the build and/or limit the amount of storage. But will this limit cause problems with not being big enough for compiling Viskores in certain contexts? If so, what would be the consequences?

kmorel:
OK. I guess that is OK. You probably know how this works better than I.

I will note that one of the CUDA builds failed to finish. That's something to fix before making build changes.

vicentebolea:
@kmorel can you review this again? Thanks :)

PR #64, opened 2025-04-16
Chase down UnitTestStreamlineMPI_mpi error

kmorel:


vicentebolea:
@kmorel we need to mark this as no draft for the Frank CI to run

kmorel:
> we need to mark this as no draft for the Frank CI to run

OK. Draft marking removed.

Is there a rational for this, or is this just the way things work?

vicentebolea:
> Is there a rational for this, or is this just the way things work?

I think that the reason for it to exist is that sometimes we might want to open a PR for the sake of a discussion but not using the precious CI. I can see how this might be useful in very large project with limited CI resources such as Spack.

In any case, it does not really makes sense to have it enabled in Viskores at this moment, fortunately this is a optional feature of the script that can be turned off. I will follow up with this change.

kmorel:
Here is an example of where the test seems to fail:

* https://open.cdash.org/tests/2048700812
* https://gitlab.spack.io/viskores/viskores/-/jobs/16866579

PR #65, opened 2025-04-18
ci: use external gl-gh-sync/github-sync@latest

vicentebolea:
This is part of an effort to common components in DAV_SDK projects which eases the maintainability of those. 

Here is an instance of it running in the Spack CI instance: https://gitlab.spack.io/viskores/viskores/-/jobs/16317691


kmorel (approved):
Looks good as long as CI still passes.

BTW, as I write this, the Frank CI check says:

> Waiting for status to be reported — waiting for format check to succeed

What does this mean? What is the check waiting for before it sends it to the Frank CI?

vicentebolea:
> What does this mean? What is the check waiting for before it sends it to the Frank CI?

The script runs every 10 mins, it just did not have enough time to run twice. Note that these changes needs to be in master to be applied.

PR #75, opened 2025-05-06
Draft: Update thirdparties

vicentebolea:
We need this thirdparty updates to get around some issues with formatting and whitespaces that blocks us to get VTK to include Viskores.


vicentebolea:
@kmorel there are no changes yet, these changes are supposed to be whitespaces issues that we are trying to address using the gitattributes, however, it does not seem to work. I accidentally marked this as ready as opposed to draft, I will close this for the time being until resolve it to avoid further confusion.

PR #80, opened 2025-05-07
Remove some CI warnings

kmorel:


kmorel:
@vicentebolea Please review. I am trying to get rid of some of the warnings we are seeing. Here is one for a -1 index we shouldn't ever encounter.

vicentebolea (approved):
Looks good!

PR #81, opened 2025-05-08
Fix compiler warnings about constructor templates

kmorel:
It is not valid to add template parameters to a constructor definition. This functionality will be removed in C++20. Remove an instance of that.


vicentebolea (approved):
Looks great! and thanks for adding the sample release note.

Issue #87, opened 2025-05-12
Loguru warning not properly suppressed


kmorel:
On exactly one dashboard, ubuntu2204_hip_kokkos43, we are seeing warnings like the following.

```
../viskores/thirdparty/loguru/viskoresloguru/loguru.cpp:1174:4: warning: 'snprintf' will always be truncated; specified size is 5, but format string expands to at least 6 [-Wformat-truncation]
```

(See for example https://open.cdash.org/sites/131711)

What I don't understand is that there is a warning exception that suppress this warning from being reported.

https://github.com/Viskores/viskores/blob/abd2519e541c823ee816c70549e9b33ba2e3ddd4/CTestCustom.cmake.in#L53-L57

This warning happens on other CI and gets properly suppressed, so why is it still reported here?

kmorel:
@vicentebolea Do you have any idea why CMake is still reporting this warning?

PR #90, opened 2025-05-12
ci: address loguru warning

vicentebolea:
fixes: #87


kmorel (approved):
Thanks for fixing this.

How does this fix the issue? Was that dashboard not doing a regex match for some reason? Does the moving of the `configure_file` command affect this?

Issue #94, opened 2025-05-14
example code in README.md uses VISKORES_DEPRECATED methods


BenWibking:
When I build the example code included in README.md, I get warnings that it uses deprecated methods:
```
[1/2] Building CXX object CMakeFiles/Demo.dir/Demo.cxx.o
/Users/benwibking/amrex_codes/viskores_example/Demo.cxx:23:17: warning: 'Tangle' is deprecated: Use SetCellDimensions or SetPointDimensions. Deprecated in version 2.0. [-Wdeprecated-declarations]
   23 |   auto tangle = viskores::source::Tangle(viskores::Id3{ 50, 50, 50 });
      |                 ^
/Users/benwibking/amrex_codes/viskores/viskores/source/Tangle.h:49:3: note: 'Tangle' has been explicitly marked deprecated here
   49 |   VISKORES_DEPRECATED(2.0, "Use SetCellDimensions or SetPointDimensions.")
      |   ^
/Users/benwibking/amrex_codes/viskores/viskores/Deprecated.h:153:18: note: expanded from macro 'VISKORES_DEPRECATED'
  153 |   __attribute__((deprecated(VISKORES_DEPRECATED_MAKE_MESSAGE(__VA_ARGS__))))
      |                  ^
1 warning generated.
[2/2] Linking CXX executable Demo
```

kmorel:
Thanks for pointing this out. The code copied to the README text file obviously did not keep up with the demo in the examples.

I submitted a PR that removes the listing directly in the README and replaces it with references to the same code in the examples. That way we can worry less about the code becoming out of date.

PR #95, opened 2025-05-15
Change the README to reference an example

kmorel:
Previously, the README.md file listed a simple example to demonstrate what Viskores code looks like. However, it was pointed out that the code was out of date (using deprecated features).

To prevent the example getting out of date, change the listing to a reference to the code in the examples directory. When viewing the README on GitHub, users will be able to simply click the link and view this file directly. Because the code is compiled as part of the CI, we don't have to worry about it being out of date.

Fixes: #94


vicentebolea (approved):
Sounds good, one less thing to update (and often forget).

All the references are correct.

PR #96, opened 2025-05-15
ViskoresCPUVectorization.cmake: fix condition for PowerPC optimizations

barracuda156:
This is not endian-specific, Big-endian PowerPC also does not support `-march=`.


vicentebolea (approved):
ppc will match ppc64le anyways. Sounds good! thanks for the contribution.

PR #100, opened 2025-05-19
Added a template constructor to MaskSelect

kmorel:
The `viskores::worklet::MaskSelect` constructor takes a mask array and constructs a lookup array for the mask operation. To speed up compilation and reduce executable sizes, the building on the mask array is pre-compiled for expected array types. However, it might be that you would like to construct a mask from a selection array with other types such as `viskores::cont::ArrayHandleTransform`. To do this efficiently, you need to compile a special version of the mask build function. To allow this, a `MaskSelectTemplate` class is added that has a templated version of the constructor.

### DIFF
--- a/docs/changelog/mask-select-template.md
+++ b/docs/changelog/mask-select-template.md
@@ -1,0 +1,11 @@
+## Added a template constructor to MaskSelect
+
+The `viskores::worklet::MaskSelect` constructor takes a mask array and
+constructs a lookup array for the mask operation. To speed up compilation and
+reduce executable sizes, the building on the mask array is pre-compiled for
+expected array types. However, it might be that you would like to construct a
+mask from a selection array with other types such as
+`viskores::cont::ArrayHandleTransform`. To do this efficiently, you need to
+compile a special version of the mask build function. To allow this, a
+`MaskSelectTemplate` class is added that has a templated version of the
+constructor.
--- a/viskores/worklet/CMakeLists.txt
+++ b/viskores/worklet/CMakeLists.txt
@@ -33,6 +33,7 @@ set(headers
   MaskIndices.h
   MaskNone.h
   MaskSelect.h
+  MaskSelectTemplate.h
   NDimsHistMarginalization.h
   Normalize.h
   ScalarsToColors.h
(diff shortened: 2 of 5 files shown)


spyridon97:
> class VISKORES_ALWAYS_EXPORT MaskSelectTemplate : public viskores::worklet::MaskSelect
> {
> public:
> };

I don't understand the above. I was expecting a class in which i pass the array handle type and it does the same as maskSelect.

spyridon97:
> class VISKORES_ALWAYS_EXPORT MaskSelectTemplate : public viskores::worklet::MaskSelect
> {
> public:
> };

I don't understand the above. I was expecting a class in in which i pass the array handle type and it does the same as maskSelect.

I would also suggest to deprecate the old mask select constructor with the unkwown array handle since it can fail and not be able to do something about.

This way MaskSelect now becomes an instatiation of MaskSelectTemplate.

kmorel:
> I don't understand the above. I was expecting a MaskSelectTemplate class with a template parameter in which i pass the array handle type and it does the same as maskSelect.

The template is in the constructor. You can give the constructor any type of `ArrayHandle`, and it will build the index mapping array (thread to output). Once the indexing array is creating, the mask no longer needs the select array, which is why the class itself does not need to be templated.

spyridon97:
> > I don't understand the above. I was expecting a MaskSelectTemplate class with a template parameter in which i pass the array handle type and it does the same as maskSelect.
> 
> The template is in the constructor. You can give the constructor any type of `ArrayHandle`, and it will build the index mapping array (thread to output). Once the indexing array is creating, the mask no longer needs the select array, which is why the class itself does not need to be templated.

By constructor you mean MaskSelectBuild?

kmorel:
> > > I don't understand the above. I was expecting a MaskSelectTemplate class with a template parameter in which i pass the array handle type and it does the same as maskSelect.
> > 
> > 
> > The template is in the constructor. You can give the constructor any type of `ArrayHandle`, and it will build the index mapping array (thread to output). Once the indexing array is creating, the mask no longer needs the select array, which is why the class itself does not need to be templated.
> 
> By constructor you mean MaskSelectBuild?

Whoops. I forgot to put in the actual constructor. I've added the constructor now and changed the code to make sure that it is actually used.

spyridon97:
+2

spyridon97 (approved):
Everything looks good now.

PR #101, opened 2025-05-20
Update use of Actor rendering class

kmorel:
When the rendering classes were originally created, `Actor` accepted the cell set, coordinate system, and field independently. More recently (but not that recently), new constructors were made that accept a `DataSet` object and pulls these out for you. This is generally a better way to construct actors, so update the documentation and testing code to follow this convention.


kmorel:
@dpugmire I usually lean on you for the rendering code in Viskores. Can you do a review of these changes. The lines are mostly documentation and style. I also added some new constructors to `Actor` that do not require specifying the coordinate system.

dpugmire (approved):
Looks good.

Issue #98, opened 2025-05-15
exception "cell set type not currently supported" when volume rendering AMR dataset


BenWibking:
I am trying to do a 3D volume rendering of the included `viskores::source::Amr` dataset using `viskores::rendering::MapperVolume`.

However, I get this runtime error:
```
libc++abi: terminating due to uncaught exception of type viskores::cont::ErrorBadValue: Mapper volume: cell set type not currently supported

Loguru caught a signal: SIGABRT
Stack trace:
16         0x18c4b2b4c start + 6000
15         0x10419fad8 main + 1028
14         0x10701419c viskores::rendering::View3D::Paint() + 124
13         0x1070072a4 viskores::rendering::Scene::Render(viskores::rendering::Mapper&, viskores::rendering::Canvas&, viskores::rendering::Camera const&) const + 120
12         0x106fd0f00 viskores::rendering::Actor::Render(viskores::rendering::Mapper&, viskores::rendering::Canvas&, viskores::rendering::Camera const&) const + 232
11         0x106ff975c viskores::rendering::Mapper::RenderCellsPartitioned(viskores::cont::PartitionedDataSet, std::string, viskores::cont::ColorTable const&, viskores::rendering::Camera const&, viskores::Range const&) + 828
10         0x106ff9414 viskores::rendering::Mapper::RenderCells(viskores::cont::UnknownCellSet const&, viskores::cont::CoordinateSystem const&, viskores::cont::Field const&, viskores::cont::ColorTable const&, viskores::rendering::Camera const&, viskores::Range const&, viskores::cont::Field const&) + 88
9          0x10708ee84 viskores::rendering::MapperVolume::RenderCellsImpl(viskores::cont::UnknownCellSet const&, viskores::cont::CoordinateSystem const&, viskores::cont::Field const&, viskores::cont::ColorTable const&, viskores::rendering::Camera const&, viskores::Range const&, viskores::cont::Field const&) + 300
8          0x18c809bf0 __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 0
7          0x18c809c48 __cxa_get_exception_ptr + 0
6          0x18c8066b0 std::__terminate(void (*)()) + 16
5          0x18c47cd84 _objc_terminate() + 172
4          0x18c7f5cf0 demangling_terminate_handler() + 316
3          0x18c80739c __cxxabiv1::__aligned_malloc_with_fallback(unsigned long) + 0
2          0x18c75ac60 abort + 124
[...]

PR #102, opened 2025-05-21
Fixed bug with `MapperVolume` blending

kmorel:
When writing cast rays back to a frame buffer, the canvas records both the color and the depth. The depth is computed from where the ray intersected the data. However, `MapperVolume` does not record this intersection depth because it traces through the whole transparent volume. This was causing bogus depth to be written to the buffer, and that was interfering when rendering multiple partitions.

This problem is fixed by adding an option to the `Canvas::WriteToCanvas` method that prevents the depth buffer from being updated. The `MapperVolume` uses this option to prevent writing bad depth values.

### DIFF
--- a/docs/changelog/mapper-volume-blend.md
+++ b/docs/changelog/mapper-volume-blend.md
@@ -1,0 +1,12 @@
+## Fixed bug with `MapperVolume` blending
+
+When writing cast rays back to a frame buffer, the canvas records both the color
+and the depth. The depth is computed from where the ray intersected the data.
+However, `MapperVolume` does not record this intersection depth because it
+traces through the whole transparent volume. This was causing bogus depth to be
+written to the buffer, and that was interfering when rendering multiple
+partitions.
+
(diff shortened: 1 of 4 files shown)


kmorel:
> Would it be possible to add a render test?
> 
> We are planning on using this in production for our code on Frontier, with code based on this proof-of-concept: https://github.com/BenWibking/viskores-miniapp/blob/1587773702fe3b98b0924b4eadce8cdd7bf2b56f/volrender.cpp.

Yes. This fixed happened because I am working on adding an AMR volume rendering test. However, that is touching more code than I like in a single PR. It relies changes on this PR and #101. Once these two PR's are merged, I can add an AMR rendering test.

Given that, @BenWibking can you provide a review?

kmorel:
Based on @BenWibking's comment on https://github.com/Viskores/viskores/issues/98#issuecomment-2906307662, I will take that as a positive review and merge.

PR #103, opened 2025-05-21
ci: correct now case-sensitive project id

vicentebolea:


vicentebolea:
@kmorel the issue was that cdash project is now case sensitive, this should fix it.

vicentebolea:

PR #109, opened 2025-06-05
Ray Culling Bug with Z Extents

nicolemarsaglia:
In the raytracing camera, remove the z extents check that can erroneously cull rays for large datasets if all corners are outside the NP & FP but there is data present within NP & FP.

This solves a bug I've been encountering of rays being culled. Here is an example where the field data was not being rendered due to ray culling, but we were still rendering the mesh, and then the fixed image. 

Not sure of the consequences, curious to see what the CI does with the unit tests. 
render_slice_mesh_and_field2
render_slice_mesh_and_field3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,8 @@ jobs:
       image: ghcr.io/viskores/format:20250212
     steps:
       - uses: actions/checkout@v4
+      - name: Mark repository as safe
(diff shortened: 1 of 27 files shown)


kmorel (approved):
These changes look good as long as they pass correctness and performance tests.

Before merging this in, please rebase onto the latest `main` branch and verify the CI again. The CI on this commit seems to have gone a little wonky.

PR #111, opened 2025-06-05
Correct issue with CI config

kmorel:
The CI configuration used the option `ECHO_ERROR_VARIABLE` for `execute_process`. However, this option is only available for CMake 3.18 and greater, and some of the CI runs as early as 3.13.5.

### DIFF
--- a/.gitlab/ci/config/initial_config.cmake
+++ b/.gitlab/ci/config/initial_config.cmake
@@ -141,7 +141,6 @@ foreach(option IN LISTS options)
     execute_process(
       COMMAND ${CCACHE_COMMAND} "--version"
       OUTPUT_VARIABLE CCACHE_VERSION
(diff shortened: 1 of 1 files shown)


.gitlab/ci/config/initial_config.cmake
```diff
@@ -141,7 +141,7 @@ foreach(option IN LISTS options)
     execute_process(
       COMMAND ${CCACHE_COMMAND} "--version"
       OUTPUT_VARIABLE CCACHE_VERSION
-      ECHO_ERROR_VARIABLE
+      #ECHO_ERROR_VARIABLE
```
vicentebolea, line 144:
```suggestion
```

vicentebolea (approved):
We can added it back if it is missed.

PR #116, opened 2025-06-13
Update widows CI to Windows Server 2022

kmorel:
Windows Server 2019 is being deprecated, so we need to update.


vicentebolea (approved):
Looks good!

PR #127, opened 2025-06-25
Add testing for rendering data with alternate representations of coordinate systems

### LINKED ISSUE #115: Problem with rendering strided arrays
Ascent is reporting problems with rendering data with strided arrays. (This is particularly for the sphere mapper but is probably similar elsewhere.) In this particular case, the data is being loaded from a funny 5-tuple and then stored in an `ArrayHandleSOA` (I think). This structure is causing cast-and-call problems further down.

The rendering code is probably not considering non-basic arrays or certainly not the standard arrays in the multiplexer. There should be a fallback for extracting components.

This problem was reported by Ascent here: https://github.com/Alpine-DAV/ascent/issues/1468.

kmorel:
There were some issues in Ascent with rendering data whose coordinates were defined as strided arrays from coordinates packed with fields (see #115 and https://github.com/Alpine-DAV/ascent/issues/1468). These issues should have been fixed in PRs #124 and #126. This addition adds a rendering regression test to make sure that this behavior remains. The rendering test will fail before the merge of these two PRs but should succeed after them.

Fixes #115

### DIFF
--- a/viskores/rendering/testing/CMakeLists.txt
+++ b/viskores/rendering/testing/CMakeLists.txt
@@ -19,6 +19,7 @@
 viskores_declare_headers(${headers})
 
 set(unit_tests
+  RenderTestAlternateCoordinates.cxx
   UnitTestCanvas.cxx
   UnitTestMapperConnectivity.cxx
   UnitTestMultiMapper.cxx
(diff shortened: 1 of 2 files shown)


nicolemarsaglia (approved):
+1 

Thanks for this!

PR #129, opened 2025-06-25
Include cstdint header for the intptr_t type

lahwaacz:
Otherwise GCC 15 complains:

```
/build/viskores/src/viskores-1.0.0/viskores/thirdparty/diy/viskoresdiy/include/viskoresdiy/thirdparty/itlib/small_vector.hpp:204:22: error: ‘intptr_t’ does not name a type [-Wtemplate-body]
  204 |     static constexpr intptr_t revert_to_static_size = RevertToStaticSize;
      |                      ^~~~~~~~
```

### DIFF
--- a/viskores/thirdparty/diy/viskoresdiy/include/viskoresdiy/thirdparty/itlib/small_vector.hpp
+++ b/viskores/thirdparty/diy/viskoresdiy/include/viskoresdiy/thirdparty/itlib/small_vector.hpp
@@ -140,6 +140,7 @@
 
 #include <type_traits>
 #include <cstddef>
+#include <cstdint>
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


vicentebolea (approved):
I have been just fixing the same issue in another project

kmorel:
@lahwaacz Thanks for bringing this issue and the fix to our attention. Unfortunately, because this is in a third-party library, we have a separate process for updates that is more complex than making the change directly in this repo. I won't go into details, but there is a third repo that maintains patches we make to the DIY library so that our patches don't get destroyed every time we update from the main project. @vicentebolea will pull your patch over to that repo and pull that change back here.

vicentebolea (approved):

PR #133, opened 2025-06-26
Fix warnings from gcc14

kmorel:


vicentebolea (approved):

kmorel:
@vicentebolea Now that the Viskores CI should build without any compiler warnings, should we change the CI setup to fail on compiler warnings?

PR #134, opened 2025-06-27
ci: set jobs to fail with any build warning

vicentebolea:
This PR will make the jobs to fail if they detect any warnings. Note that it
will not cancel the build, instead, once finished, it will mark it as failure.


kmorel (approved):
I just love seeing that green dashboard! 😃 

I'm going to assume this will report a failed check on a warning. We'll find out soon enough if it doesn't.

PR #137, opened 2025-07-15
Use Project Name Instead of Viskores_BINARY_DIR to Detect Downstream Usage

qbisi:
Viskores_BINARY_DIR is not a reliable variable for determining whether the current project is the main project or a downstream one. In environments like Nix/Nixpkgs, everything is built in a sandbox, and Viskores_BINARY_DIR is typically /build/source/build/. Instead, use the project name to make this distinction.


vicentebolea (approved):
Thanks for the contribution. It makes sense.

PR #139, opened 2025-07-17
Fix bug in celllocator

dpugmire:


kmorel:
You need to back out of the ["restart tests." commit](https://github.com/Viskores/viskores/pull/139/commits/7c2a3c8f6e3d4c4e844a27c7432619d14f841c60). It reverted the data submodule, which is causing tests to fail.

kmorel:
BTW, if you just want to kick a new set of tests, you can "touch" the most recent commit by changing the date.

```sh
git commit --amend --no-edit --date "`date`"
```

The advantage of that is you don't have to make a pointless edit to a file. Git and the test framework will register a new commit that needs to be retested even though none of the files changed. Of course, since you are changing the commit, you have to force the push because you are replacing to most recent commit on the GitHub branch.

```sh
git commit --amend --no-edit --date "`date`"
git push -f
```

I can never remember this touch command, so I add it as an alias with configuration like this in my `~/.gitconfig` file.

```ini
[alias]
        touch = !sh -c 'git commit --amend --no-edit --date \"`date`\"'
```

If you have that, then the commands to restart the tests would be

```sh
git touch
git push -f
```

One day we should have a bot that allows you to selectively restart tests by adding GitHub comments so we don't have to do these silly tricks.

PR #146, opened 2025-08-06
Streamline licensing

pombredanne:
Some suggestion to streamline licensing documentation.

- Reference: https://github.com/aboutcode-org/scancode-toolkit/issues/4507

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,19 +1,6 @@
 ##============================================================================
-##  The contents of this file are covered by the Viskores license. See
-##  LICENSE.txt for details.
-##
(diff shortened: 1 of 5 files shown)


LICENSE.txt
```diff
@@ -27,14 +27,14 @@ met:
    documentation and/or other materials provided with the
    distribution.
 
- * Neither the name of Kitware nor the names of any contributors may
+ * Neither the name of the copyright holder nor the names of its contributors may
```
pombredanne, line 30: Note that this is a set of minimal changes to minimize diff. I did not use the numbers from each clause from the OSI text.

PR #148, opened 2025-08-21
ci: add no_testing cuda13 job

vicentebolea:
We cannot add the testing phase yet since the UO Frank cluster does not
    have a recent enough nvidia driver.
    
    This commit also rename the Frank Cluster cuda builds to include
    in their name the cuda version.

cc @spyridon97

### DIFF
--- a/viskores/Swap.h
+++ b/viskores/Swap.h
@@ -42,8 +42,18 @@ namespace viskores
 // `Swap`, it results in ADL being used, causing the templated functions `cub::Swap` and
 // `viskores::Swap` to conflict.
 #if defined(VISKORES_CUDA_VERSION_MAJOR) && (VISKORES_CUDA_VERSION_MAJOR >= 12)
+#if VISKORES_CUDA_VERSION_MAJOR == 12
 using cub::Swap;
+#elif VISKORES_CUDA_VERSION_MAJOR >= 13
+template <typename T>
+VISKORES_EXEC_CONT inline void Swap(T& a, T& b)
+{
+  using ::cuda::std::swap;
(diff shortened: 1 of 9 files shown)


viskores/Swap.h
```diff
@@ -42,8 +42,24 @@ namespace viskores
...
+  using std::swap;
+  swap(a, b);
+}
+template <typename T>
+__device__ inline void Swap(T& a, T& b)
+{
+  using ::cuda::std::swap;
+  swap(a, b);
+}
+#endif
```
kmorel, line 60: I am concerned about this change. As I recall from another thread, the problem this is trying to get around is that `cub::Swap` is deprecated in Cuda 13. However, even though it is deprecated, it still exists. The whole reason the `using cub::Swap` statement exists is to prevent ambiguous references to `Swap` when it is used in a templated function.

You may need to just allow the warnings until `cub::Swap` is completely removed.

vicentebolea: The problem is that cuda 13.0 seems to have effectively dropped cub::Swap as shown here: https://github.com/NVIDIA/cccl/commit/d090fb6eb81b21791fb4655ddf2926d0af1eb4b6

This is now provided under cuda::std::Swap as shown here: https://github.com/NVIDIA/cccl/blob/main/libcudacxx/include/cuda/std/__utility/swap.h

kmorel: OK. As long as `cub::Swap` as been fully dropped, then this replacement is OK.

kmorel (approved):
Looks good as long as the CI passes.

PR #154, opened 2025-08-25
Update to 1.1.0

vicentebolea:
We need to backport HEAD~1 to main after merge so that Viskores main remains at the 1.1.X version


vicentebolea:
> Is this just merging in the latest `main` into `release` to update to 1.1? That seems fine with me.
> 
> I assume a `release-1-0` branch will be created to support backports and patch releases.
> 
> I also assume there is still work to set up the RC such as building the release notes.

Yes, we are merging latest main into release, however we will not bring all the release changes to main. (last commit). I have crafted a tentative release notes files that we can cure until the release the final release.

Thanks for remainding, yes I will push a new release-1.0 based on the release branch commit ancestor before this merge commit.

PR #161, opened 2025-08-30
Make the device library private

kmorel:
Viskores no longer makes the device libraries public on its exported `viskores::cont` target. Previously, the device libraries were linked to `viskores::cont` as `PUBLIC`, which means that pretty much anything that used Viskores also brought in any compiler flags for the device. This sometimes forced downstream code to use a device compiler when it did not need to.


kmorel:
@vicentebolea Can you take a look at the testing errors that are happening on [ubuntu2204_kokkos37_cuda122](https://gitlab.spack.io/viskores/viskores/-/jobs/18059281)? I can't figure out what is going on. Most of the tests are seg faulting on program exit and then locking up until the test times out.

I didn't make any code changes that should cause that. My best guess is that my changes to the linking has changed the order of the libraries in the linker (or perhaps replicated some), and that is causing some weird behavior. Perhaps the different order is causing issues in the shutdown sequence and causing something in Kokkos to try to clean up after CUDA or something else important shuts down. (That said, Kokkos shuts down pretty early.)

kmorel:
@vicentebolea Updating to Kokkos 4.1 has not helped. I'm still pulling my hair out on this.

Do you have any suggestions on why changing  the link dependencies causes these malloc issues?

It could be a buffer overrun that trashes the malloc state. But I can't get any sanitizer to report any issues with buffer allocations or overruns.

Could there be an issue with Kokkos compiling to static libraries? Could the object code be replicated in the Viskores shared libraries and cause some duplicated free somehow?

vicentebolea:
@kmorel I am not sure about the reason of it, ill return from vacations in 10 days. Ill revisit this then if the reason is not found ye.

kmorel:
@vicentebolea Last I checked, this was still an issue. Do you have any further thoughts on what might be wrong?

.gitlab/ci/docker/ubuntu2204_kokkos_cuda.dockerfile
```diff
@@ -53,14 +53,16 @@ RUN mkdir -p /opt/kokkos/build && \
     mkdir bld && cd bld && \
     CXX=/opt/kokkos/build/kokkos-$KOKKOS_VERSION/bin/nvcc_wrapper \
     cmake -B . -S ../kokkos-$KOKKOS_VERSION \
+          -DBUILD_SHARED_LIBS=ON \
```
vicentebolea: Kokkos sets BUILD_SHARED_LIBS=OFF by default (as in cmake's default behavior)

kmorel: It's weird that Kokkos compiles OK but this breaks our code.

Do you think the `-DCMAKE_CXX_FLAGS=-fPIC` is actually the problematic one?

PR #163, opened 2025-09-01
Require initialize

kmorel:
Update viskores to require calling `viskores::cont::Initialize`.

This is currently a soft requirement, but warnings will happen if you do not
call it.


vicentebolea (approved):
Clever way to keep state, looks good!

PR #166, opened 2025-09-09
Consolidate testing

kmorel:
The testing helper code in Viskores has been consolidated. Originally, there was a separation between the tests that used code in the control environment and that which did not. This was originally from back in the day when Viskores (or rather, its predecessor) was envisioned as a header-only library. However, this separation is no longer meaningful. In fact, the testing code without the control environment was missing some initialization that can cause issues with some devices.

Now, all the testing helper code is accessible from `viskores/testing/Testing.h`. The old `viskores/cont/testing/Testing.h` still exists (for now), but it is an empty file pointing back to `viskores/testing/Testing.h`.


kmorel:
This change is required for #161.

kmorel:
@vicentebolea Although this looks like a large change, it is really just moving the code from `viskores/cont/testing/Testing.h` to `viskores/testing/Testing.h`.

I've been thinking about doing this for a long time. The changes for #161 seem to break the behavior of not calling `Initialize` for some Kokkos features, so the tests that were avoiding using the cont library were breaking.

PR #182, opened 2025-09-28
Improve compilation of flow filters

kmorel:
Several of the flow filters contain multiple flow paths through different
worklets. This can sometimes overwhelm device compilers. To improve compilation,
the compilation of worklets for the flow filters is separated into different
translation units using the Viskores instantiation compile feature. This reduces
the burden on any particular use of a compiler and helps leverage parallel
compiling.

### DIFF
--- a/docs/changelog/faster-flow-build.md
+++ b/docs/changelog/faster-flow-build.md
@@ -1,0 +1,8 @@
+## Improve compilation of flow filters
+
+Several of the flow filters contain multiple flow paths through different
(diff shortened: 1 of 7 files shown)


dpugmire (approved):
looks good.

spyridon97:
Consider adding instantiations for Probe, MeshQuality, and MarchingCells also.

kmorel:
Backport: #206 

Are there any specific filter under mesh quality that is problematic? The code is already broken up quite a bit in different filters.

Issue #183, opened 2025-10-08
Failure of UnitTestBoundingIntervalHierarchy


dpugmire:
PR #136 introduced a CUDA-related bug that causes intermittent test failures.
This issue needs to be investigated and resolved.
In the meantime, I’m disabling the affected test to keep the dashboards clean.

kmorel:
Fixed in #270.

PR #184, opened 2025-10-08
Disable CellLocatorBoundingIntervalHierarchy

dpugmire:
Issue #183 describes a periodic failure in the unit tests for CellLocatorBoundingIntervalHierarchy. Disabling this test for now while the bug is investigated.


kmorel:
If we are disabling this test, we should also deprecate `CellLocatorBoundingIntervalHierarchy`. There appears to be a problem with it, and I don't think anyone has plans to figure out and fix the issue. Thus, it makes sense to discourage its use and eventually remove the class.

spyridon97:
Deprecating BIH Cell Locator is in my opinion a bad idea given that they are the fastest for non-uniform meshes (at least that's the case for VTK). What seems to be the problem?

PR #195, opened 2025-10-22
Support BSD sed extensions in transition script

kmorel:
This includes replacing `[[:<:]]` and `[[:>:]]` for `\b`. Also note that the BSD word boundaries do not consider `_` a word boundary, so additional checks for those are added as well.

### DIFF
--- a/Utilities/Scripts/vtkm_to_viskores_file.bash
+++ b/Utilities/Scripts/vtkm_to_viskores_file.bash
@@ -8,11 +8,32 @@
 ##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
 ##============================================================================
 
+if [ $(echo in | sed -E 's/\b(in)\b/out/') = out ] ; then
+  # GNU extensions
+  wb='\b'
+  we='\b'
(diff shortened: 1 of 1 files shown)


Utilities/Scripts/vtkm_to_viskores_file.bash
```diff
@@ -8,11 +8,32 @@
 ##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
 ##============================================================================
 
+if [ $(echo in | sed -E 's/\b(in)\b/out/') = out ] ; then
+  # GNU extensions
+  wb='\b'
+  we='\b'
+elif [ $(echo in | sed -E 's/[[:<:]](in)[[:>:]]/out/') = out ] ; then
+  # BSD extensions
+  wb='[[:<:]]'
```
vicentebolea:
```suggestion
  wb='([[:<:]]|_)'
```

Something like this could help remove the unrolled variations of the sed command invocation below

kmorel: I don't think that will work. If you add that, the regex will match the `_` and the whole `_vtkm` will be replaced with `viskores` thereby deleting the `_`.

You might be able to get clever with group substitutions, but that gets tricky.

PR #199, opened 2025-10-24
Simplify code for ArrayHandleSOA portal

kmorel:
The ArrayPortalSOA had some clever template code to set and get values in multiple arrays. However, we have found that some device compilers get bogged down resolving all these template parameters in kernels, so the code has been significantly simplified.

### DIFF
--- a/viskores/cont/ArrayExtractComponent.h
+++ b/viskores/cont/ArrayExtractComponent.h
@@ -52,7 +52,7 @@ ArrayExtractComponentFallback(const viskores::cont::ArrayHandle<T, S>& src,
 {
   if (allowCopy != viskores::CopyFlag::On)
   {
-    throw viskores::cont::ErrorBadValue(
+    throw viskores::cont::ErrorBadType(
       "Cannot extract component of " +
       viskores::cont::TypeToString<viskores::cont::ArrayHandle<T, S>>() + " without copying");
   }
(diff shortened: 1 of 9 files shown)


spyridon97:
I tried 2 times, both times got killed at around 30 minutes.

kmorel:
Thanks for continuing to pound on this.

I wouldn't consider the problem "fixed" even if we live with a 2-hour compile. This change was just to give us some information. The compiles seem to be using the scalar version, which adds some overhead that drives the compile to fail. However, even when compiling without the scalars, something is still causing a slowdown.

Perhaps the extra memory in the `ArrayHandleSOAStride` portal is causing a problem? It adds 192 bits per component. Maybe that causes a layer of register spilling that is problematic for the compiler?

I just added another commit that swaps the SOA stride portal for the basic SOA portal. This will not run correctly, but can you test the compile for that?

kmorel:
OK, _that_ is unexpected. I made the code more complex since last time, not less.

I realized that I left some invalid code in my last commit that could invalidly speed up copied. Let's try another pull/compile.

spyridon97:
it compiled again in 45 minutes.

spyridon97:
it compiled successfully in 51 minutes.

kmorel:
> it compiled successfully in 51 minutes. If the commits are cleaned-up, let me know so that i can review them, also let's make sure this is backported into release.

I cleaned up the commits. I also corrected some problems that popped up by removing the div/mod. (Unfortunately, `ArrayHandleConstant` is one of the arrays that lost support.)

You may want one more test compile, but if the CI passes I think we are good. If that works, I will cherry-pick this back to the release branch.

PR #201, opened 2025-10-31
Improve testing for ArrayHandleSOAStride

kmorel:
Previously, the divisor and modulo were not tested. They are now.

### DIFF
--- a/viskores/cont/testing/UnitTestArrayHandleSOAStride.cxx
+++ b/viskores/cont/testing/UnitTestArrayHandleSOAStride.cxx
@@ -19,7 +19,9 @@
 #include <viskores/cont/ArrayHandleSOAStride.h>
 
 #include <viskores/cont/ArrayCopyDevice.h>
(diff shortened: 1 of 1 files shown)


vicentebolea (commented):
I cannot see how constructing a SOA array from a AOS and checking that its is equal to its source tests div and mod. Maybe I am missing something or this is a very cool trick 😄

kmorel:
The test comes from constructing an `ArrayHandleSOAStride` from an `ArrayHandleUniformPointCoordinates`. This works by constructing an array for each axis (in the same way that `ArrayHandleCartesianProduct` works). The arrays are then accessed through divs and mods to expand each array. For example, the x array is generated with a mod by the x-size. The y array is generated by a div of the x-size and mod of the y-size.

vicentebolea:
Oh I understand now,I should have looked at the `ArrayHandleSOAStride` implementation before asking. Thanks for explaining it.

PR #203, opened 2025-10-31
Fix ANARIMapperGlyphs dispatching

kmorel:
The dispatching caused issues when using things that are not Vec-3, which can happen if the default types include other vec types.


kmorel:
@jeffamstutz Can you do a review of these changes that fix some type conversion issues under certain compile configurations?

PR #207, opened 2025-11-05
Initial ANARI device

kmorel:
Viskores now provides an ANARI device that can be used by applications that use
ANARI for rendering. The intention of this device is to provide a simple,
portable, accelerated ANARI device that will be available in HPC systems
regardless of vendor support. We hope this will help jumpstart the support of
ANARI for scientific visualization applications at HPC centers.

This device is still in its experimental phase. Although functional, it is
missing many features that applications will expect to be supported. The
inclusion of the device into Viskores in this early phase should help promote
development and simplify the integration of any changes necessary in the
rendering library.

### DIFF
--- a/CMake/ViskoresCheckLicense.cmake
+++ b/CMake/ViskoresCheckLicense.cmake
@@ -46,6 +46,9 @@ set(EXCEPTIONS
   DCO.txt
   README.txt
   docs/users-guide/requirements.txt
+  # Common directories with build files
+  .venv
+  build
   )
 
 if (NOT Viskores_SOURCE_DIR)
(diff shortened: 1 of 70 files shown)


jeffamstutz:
@berkgeveci FYI

jeffamstutz:
I can have the SDK release done before the end of this week -- we have new Python bindings in the SDK that I wanted to include which got merged yesterday. No major blockers over there, so this is a good reason to prioritize that.

vicentebolea:
> Great.
> 
> @vicentebolea, Can you help me update our CI image to have ANARI 0.15? I believe this should be done on a different PR.

Sure find it here: https://github.com/Viskores/viskores/pull/210

PR #210, opened 2025-11-06
ci: update anari to 0.15.0 in ubuntu2004_kokkos image

vicentebolea:
Merge this PR before merging: https://github.com/Viskores/viskores-data/pull/4


kmorel:
@jeffamstutz The update to Anari 0.15 caused the rendering tests to detect changes in the images (see https://open.cdash.org/viewTest.php?onlyfailed&buildid=10805809). It looks like poly data got a bit brighter and volume data changed its interpolation method to make the data more transparent.

Can you verify that the differences in images are expected due to changes in the Helide device?

jeffamstutz:
All images look reasonable to me! Helide is much more "conformant" now (mostly visible with volumes), so the differences are OK to me.

kmorel:
Cool.

@vicentebolea If you update the baseline images for the ANARI tests, this PR should be ready.

kmorel (approved):

PR #213, opened 2025-11-10
Fix CMake for Kokkos-thrust (Backport #209 to Release 1.0)

kmorel:
There was a mistake in an expression that caused hip support in thrust to never be selected. This has a high performance consequence.

There was also an error in how the rocthust library was linked. It was using an invalid generate expression when linking libraries.

Backport of #209


vicentebolea (approved):
Looks good! Macos job failure is unrelated.

Issue #225, opened 2025-11-21
Adding support for more cell list types in Tube filter


DeVasconcelos:
Hello!

We are using Viskores to run visualization algorithms on the GPU in our open-source visualization software Vistle. When trying out the Tube filter, we ran into the following issue:

## Problem Description
When calling the Tube filter on a dataset with an explicit cell set with a non-basic storage type, the  following check fails in `viskores/filter/geometry_refinement/worklet/Tube.h` (lines 624 - 628):
```c++
    if (!cellset.CanConvert<viskores::cont::CellSetExplicit<>>() &&
        !cellset.CanConvert<viskores::cont::CellSetSingleType<>>())
    {
      throw viskores::cont::ErrorBadValue("Tube filter only supported for polyline data.");
    }
```

In Vistle, polyline grids are defined very similarly to explicit cell sets, they do, however, not have a shapes array. When converting from Vistle to Viskores datasets (which is necessary to use the filters), we are, thus, using the `ArrayHandleConstant` fancy array handle to create the shapes array without having to store it in memory. But since this makes the cell set an `viskores::cont::CellSetExplicit<typename viskores::cont::ArrayHandleConstant<viskores::UInt8>::StorageTag>`, we cannot pass the dataset to the Tube filter.

For clarity and more details, I've added a code example to reproduce the problem without Vistle at the end of this issue.

## Workaround

We have found the following workaround to making the Tube filter work with these types of cell sets:

### Step 1
Similar to how it is done in `viskores/cont/internal/DefaultTypesVTK.h.in`, we override the `VISKORES_DEFAULT_CELL_SET_LIST_UNSTRUCTURED` makro with a list containing the explicit cell list type with the non-basic storage type. When compiling Viskores, we set `Viskores_DEFAULT_TYPES_HEADER` to the file containing this change:
```c++
namespace viskores
{
namespace cont
{

using VistleCellSetListUnstructured = viskores::List<
    viskores::cont::CellSetExplicit<>,
    viskores::cont::CellSetExplicit<typename viskores::cont::ArrayHandleConstant<viskores::UInt8>::StorageTag>,
    viskores::cont::CellSetSingleType<>
        >;

}
}

#define VISKORES_DEFAULT_CELL_SET_LIST_UNSTRUCTURED ::viskores::cont::VistleCellSetListUnstructured
```

### Step 2
[...]

PR #228, opened 2025-11-25
Make Tube filter check all types in VISKORES_DEFAULT_CELL_SET_LIST_UNSTRUCTURED

DeVasconcelos:
This ensures the Tube filter works even if the Viskores algorithms have been compiled for use in external software. 

Together with #227 this fixes issue #225.


kmorel (approved):
Looks good. Thanks for the fix.

PR #241, opened 2026-01-12
Fix the SetupForDevelopment.sh script

kmorel:
The CONTRIBUTING.md document instructs developers to run the `Utilities/SetupForDevelopment.sh` script. However, when you ran this script, you got an error:

```bash
$ ./Utilities/SetupForDevelopment.sh
Setting up useful Git aliases...
./Utilities/SetupForDevelopment.sh: line 22: syntax error: unexpected end of file
```

This error has been fixed, and the `SetupForDevelopment.sh` script now correctly sets up git aliases.

Also added the `--recursive` argument to the submodule update part of the pullall alias. This does not currently do anything as no submodule has submodules of its own. However, adding this argument does not hurt and it might be needed in the future if a submodule is eventually added with submodules of its own. Thus, to prevent developers having to update their setup, just add the argument for now.

### DIFF
--- a/Utilities/SetupForDevelopment.sh
+++ b/Utilities/SetupForDevelopment.sh
@@ -15,6 +15,7 @@ git config rebase.stat true
 git config branch.main.rebase true
 
 # General aliases that could be global
-git config alias.pullall '!bash -c "git pull && git submodule update --init"' &&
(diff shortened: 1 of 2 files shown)


vicentebolea (approved):
Good catch! I missed this during the transition.

Issue #244, opened 2026-01-20
Add PanX and PanY


nicolemarsaglia:
While implementing tiled rendering in Ascent (https://github.com/Alpine-DAV/ascent/pull/1646), we required a viskores patch that added Pan for the Raytracer. 

https://github.com/Alpine-DAV/ascent/blob/72d03e0e3419531504e642a089032c8033ed328a/scripts/build_ascent/2026_01_02_viskores_implent_pan_raytracing.patch

[2026_01_02_viskores_implent_pan_raytracing.patch.txt](https://github.com/user-attachments/files/24749387/2026_01_02_viskores_implent_pan_raytracing.patch.txt)

PR #245, opened 2026-01-20
Implement pan with 3D camera

kmorel:
Although the `Camera` object for 3D rendering provided panning options in the image plane, the underlying camera in the raytracing code did not support it. The raycast mapping now supports X/Y panning when rendering 3D objects.

Fixes: #244


nicolemarsaglia (approved):
Thanks for adding this!

PR #250, opened 2026-01-30
add security policy file

### LINKED ISSUE #247: Add security policy
Part of the [OSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#security-policy) score is the presence and content of a security policy.  Traditionally that's a `SECURITY.md` file in the top level of the repository.

Here are a few security policies for PESO member projects to use as a starting place:
- https://github.com/HDFGroup/hdf5/blob/develop/SECURITY.md
- https://github.com/ornladios/ADIOS2/blob/master/SECURITY.md
- https://github.com/visit-dav/visit/blob/develop/SECURITY.md

vicentebolea:
fixes #247 

This PR add a Security policy file inspired by the one existing in the ADIOS2 project. This security policy filly complies with the OSSF best practices.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,28 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 1.x     | :white_check_mark: |
+
+## Reporting a Vulnerability
+
+The Viskores team and community take security bugs in Viskores seriously.
+We appreciate your efforts to responsibly disclose your findings, and will make
+every effort to acknowledge your contributions.
+
+To report a security issue, please use the GitHub Security Advisory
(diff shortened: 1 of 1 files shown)


vicentebolea:
> However, the Linux Foundation only allows members of the mailing list to post to the mailing list. So, to post a security issue, a reporter would have to join the mailing list. And even if we thought that would be an OK thing to make everyone do, that would mean that anyone who reported would see any other security issues reported, which would be a security risk.

In ADIOS2 while exploring other possible providers for the mailing list we encountered similar issues. In our Kitware google groups we have full control of its privacy controls.

I have just requested the creation of the group to sysadmin, I will add you as owner as soon as it is created.

kmorel:
Still +1

PR #251, opened 2026-02-14
Add Scorecard workflow for supply-chain security

vicentebolea:
This workflow automates Scorecard supply-chain security checks, including branch protection and scheduled analysis. This will also automatically report to our Security dashboard.


vicentebolea:
@kmorel this will run in master and scheduled once a week and will report in out security tab issues that it finds.

PR #257, opened 2026-03-11
Avoid deep copies between host and device on AMD APUs

DeVasconcelos:
Hello again!

As you might remember, we are using Viskores to run visualization algorithms on the GPU in our visualization software Vistle. Currently, we are testing the GPU modules on the supercomputer [Hunter](https://www.hlrs.de/solutions/systems/hunter) which uses AMD Instinct MI300A APUs, i.e., a special kind of accelerator where the CPUs and GPUs physically share their memory. As they are AMD devices, we are using the Kokkos+HIP build of Viskores.

## Problem Description

When profiling our test visualization workflows, we have noticed that the Kokkos device adapter always performs deep copies between the host and device, even though they are not necessary on APUs (as the device can always access the memory the host allocates and vice versa). This is in contrast to the CUDA device adapter which takes managed memory pointers into account.

## Proposed Solution

If the Kokkos flag `KOKKOS_ARCH_AMD_GFX942_APU` (this seems to be the only APU flag in Kokkos as of now) is defined, the `CopyHostToDevice` and `CopyDeviceToHost` methods use shallow instead of deep copies, similar to how this is implemented in the CUDA device adapter.

## Alternative Approach

I have also considered using `KOKKOS_HAS_SHARED_SPACE` which, unlike the very specific  `KOKKOS_ARCH_AMD_GFX942_APU`, also works for managed memory or unified virtual memory (see [this branch](https://github.com/DeVasconcelos/viskores/commits/use-kokkos-shared-space-for-unified-memory-devices/)). I did, however, abandon this approach, as it does not seem to be possible to query if a pointer was created in unified memory or not using Kokkos. 

While this alternative approach worked on the APU, I ran into illegal memory access errors when testing it on a NVIDIA GPU with UVM. I suspect this is because the allocations done on the host (unlike the ones on the device) are not controlled by Kokkos, and are thus also not created in `SharedSpace` making it inaccessible to the device. This does not happen in the CUDA device adapter as its `IsManagedPointer` method checks if a pointer is a managed pointer using  `cudaPointerGetAttributes`. Only if it is, a shallow copy is done, otherwise it performs a deep copy.

I'd be happy to hear your thoughts on this!

### DIFF
--- a/docs/changelog/avoid-copies-on-amd-apus.md
+++ b/docs/changelog/avoid-copies-on-amd-apus.md
@@ -1,4 +1,4 @@
 ## Avoid copies on AMD APUs in the Kokkos build of Viskores
 
 APUs, such as the AMD Instinct MI300A, are a special kind of accelerator where the CPU and GPU physically share their memory.
-The Kokkos build of Viskores now takes this characteristic into account by performing shallow copies between host and device if `KOKKOS_ARCH_AMD_GFX942_APU` is defined.
+The Kokkos build of Viskores now takes this characteristic into account by defining `VISKORES_PHYSICALLY_SHARED_MEMORY` and performing shallow copies between host and device, if the device is an APU.
--- a/viskores/cont/kokkos/internal/DeviceAdapterMemoryManagerKokkos.cxx
+++ b/viskores/cont/kokkos/internal/DeviceAdapterMemoryManagerKokkos.cxx
@@ -21,6 +21,12 @@
 #include <viskores/cont/kokkos/internal/KokkosAlloc.h>
 #include <viskores/cont/kokkos/internal/KokkosTypes.h>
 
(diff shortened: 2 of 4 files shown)


DeVasconcelos:
Thank you very much for the helpful feedback! I have applied the changes and tested them on the cluster. 

I did, however, place the definition of `VISKORES_PHYSICALLY_SHARED_MEMORY` into DeviceAdapterMemoryManagerKokkos.cxx instead of DeviceAdapterMemoryManager.cxx to make sure it can be found by the Kokkos memory manager and that `KOKKOS_ARCH_AMD_GFX942_APU` is available as well. Is this alright or would you prefer a more general `VISKORES_PHYSICALLY_SHARED_MEMORY`, e.g., in DeviceAdapterMemoryManager.h?

PR #258, opened 2026-03-13
Workletize streamlines

dpugmire:
Worklet-ize the flow code.


kmorel (approved):
Sorry for the delay in reviewing. This fell in my email.

Overall this looks good. I didn't do a deep dive, but the code looks reasonable.

Before this can be merged, the CI has to pass. The first problem is that it is failing the formatting test. You can get a patch file from GitHub to do that for you. (See [CONTRIBUTING](https://github.com/Viskores/viskores/blob/main/CONTRIBUTING.md#reformat-a-topic) for details). It looks like the formatting test is holding up other tests, so we will have to see if those pass.

I would also prefer that the commits get squashed before being merged.

PR #260, opened 2026-03-16
Add FindCellId and FindAllCellIds to cell locators.

dpugmire:
Add cell-id-only locator APIs to cell locator classes.

Refactored CellLocatorTwoLevel and CellLocatorUniformBins to share common traversal logic via a single implementation plus writer policies (cellId+pcoords vs cellId-only), with behavior preserved.

### DIFF
--- a/viskores/cont/CellLocatorBoundingIntervalHierarchy.h
+++ b/viskores/cont/CellLocatorBoundingIntervalHierarchy.h
@@ -35,6 +35,23 @@ namespace viskores
 namespace cont
 {
 
+namespace detail
+{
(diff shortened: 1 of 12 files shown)


kmorel (approved):
The code is fine.

I am curious about how effective this change is. I assume the point of `FindCellId` is to avoid the computation of the parametric coordinates when not needed. But the parametric coordinates are computed for unstructured grids anyway since they are needed to determine whether the point is in the cell. Do you really save a noticeable amount of time on the few arithmetic instructions saved for structured grids?

PR #261, opened 2026-03-17
Do not return early from CMakeLists.txt

### LINKED ISSUE #259: property IMPORTED_LOCATION... on TBB::tbb breaks build
656537732a42d6efafa9e8620f185973861e6d85 added the macro `_viskores_add_imported_lib_to_list` and calls it for some dependencies, e.g. `TBB::tbb` to read the property `IMPORTED_LOCATION_RELEASE`, `IMPORTED_LOCATION`, and `IMPORTED_LOCATION_DEBUG`. If one of these properties is found on a dependency, the macro extends `lib_args` and calls `return()` - which stops processing the top-level `CMakeLists.txt` and prevents Viskores from being built.

Removing the `return()` from the macro or replacing it with `break()` both solved the problem for me. I don't know if it is sufficient to handle one of these properties or if each of them should be handled.

aumuell:
Removing the return() from the macro _viskores_add_imported_lib_to_list in the top level CMakeLists.txt, as otherwise a dependency carrying the property IMPORTED_LOCATION{,_RELEASE,DEBUG would end processing immediately, so that nothing is built.

Fixes: #259


kmorel (approved):
Look good! Once the CI completes, we'll merge this in.

aumuell:
> Look good! Once the CI completes, we'll merge this in.

Thank you!

kmorel:
@vicentebolea The [smoke_test built using CMake](https://open.cdash.org/tests/2462283983) passes. They must be using the same libraries (or, rather, the same libraries are available to both builds). What is the CMake build of the smoke test doing different that makes it work?

kmorel:
@vicentebolea I can't remember if you are back this week, but when you have a chance please look in to why we are getting errors with the static make builds.

vicentebolea:
@kmorel I have just returned, I'll put this on my backlog.

vicentebolea:
superseded by: #288

kmorel:
@aumuell Your changes have been implemented and merged in #288.

Thanks for your help in fixing this, and let us know if you run into anything else.

PR #263, opened 2026-03-19
Anari raycast directly

kmorel:


jeffamstutz (approved):
With what I can understand, this LGTM. I don't have deep knowledge of `viskores::rendering`, but nothing looks out of place from my side.

+1

PR #265, opened 2026-03-19
Support skipping tests

kmorel:
Viskores tests can now skip themselves. Sometimes it is not possible to determine whether a particular test is supported until the test is run and the system can be introspected. In this case, the test can now return by invoking the `VISKORES_TEST_SKIP` macro. In this case, the test will immediately end, and CTest will report that the test was skipped rather than pass or fail.


kmorel:
@vicentebolea Ping

vicentebolea (approved):
Looks great! sorry I missed this earlier, I can see that one of the tests gets skipped.

PR #267, opened 2026-03-20
Anari test rendering

kmorel:
By default, test the ANARI interop using the built-in ANARI device.


kmorel:
I need to set up a second job with ANARI so I can test ANARI with both helide and viskores devices. That way we can make sure the viskores rendering reasonably matches the helide reference.

PR #272, opened 2026-03-24
Fix crash in multi-threaded filter execution with OpenMP backend

dpugmire:
OpenMP and TBB are not thread-safe. So disable the multiblock threading in the Filters for those devices.

### DIFF
--- a/docs/changelog/multi-block-filter-threading.md
+++ b/docs/changelog/multi-block-filter-threading.md
@@ -1,0 +1,3 @@
+## Deprecated CPU threading for multiblock filters.
+
+OpenMP and TBB backends are not threadsafe. Disable setting number of threads for these backends.
--- a/viskores/filter/Filter.cxx
+++ b/viskores/filter/Filter.cxx
@@ -109,12 +109,17 @@ viskores::cont::PartitionedDataSet Filter::DoExecutePartitions(
 {
   viskores::cont::PartitionedDataSet output;
 
-  if (this->GetRunMultiThreadedFilter())
+  const bool runMultiThreaded = this->GetRunMultiThreadedFilter();
+  const viskores::Id numThreads =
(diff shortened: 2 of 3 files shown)


kmorel (commented):
That's a good catch. I didn't think about a multi-threaded library not being thread safe.

kmorel (changes requested):
This looks good. Just two small changes:

1. Add a changelog to document the fix.
2. Since `Filter::NumThreadsPerCPU` is no longer being used, delete that member and mark `SetThreadsPerCPU` and `GetThreadsPerGPU` as deprecated.

kmorel:
It looks like [UnitTestMultiBlockFilter](https://open.cdash.org/queryTests.php?project=Viskores&filtercount=1&showfilters=1&field1=testname&compare1=61&value1=UnitTestMultiBlockFilter&date=2026-03-31)[UnitTestMultiBlockFilter](https://open.cdash.org/queryTests.php?project=Viskores&filtercount=1&showfilters=1&field1=testname&compare1=61&value1=UnitTestMultiBlockFilter&date=2026-03-31) failed on the MacOS build again. That's not great.

dpugmire:
I enabled threading for Serial device. Disabled for TBB and OpenMP.

kmorel (approved):
This seems a little redundant as it prevents multicore blocks when a multicore processor is available and then checks again, but that should be OK.

UnitTestMultiBlockFilter is passing now. However, I'm not sure it is because of these changes. It seemed to be failing everywhere yesterday, but seems to be passing elsewhere today.

dpugmire:
Good catch. I removed the redundant check.

kmorel (approved):
The changes look good (although you probably don't need the changes to `Filter.h`). Let's hope this fixes the problem.

PR #273, opened 2026-03-24
Build CI in /tmp directory

kmorel:
We have been seeing some failures on runners where build *.o files seem to disappear. That is, they must be generated because their associated compile succeeds, but the link fails because it cannot find that file.

As a quick workaround, it was suggested to try to build to the local /tmp directory instead.


kmorel:
Closing this as I clearly don't know what I'm doing.

PR #277, opened 2026-03-31
Use compliant MSVC preprocessor for device code

### LINKED ISSUE #276: viskores has a compilation error using VS2022 on CUDA 13.2
Image

kmorel:
MSVC has some behavior in their "traditional" preprocessor that is not compliant with the C99 and C++11 standards. Starting in MSVC 19 (the current earliest version supported by Viskores), MSVC provides a "compliant" preprocessor that does meet C and C++ standards. This is enabled with the `/Zc:preprocessor` command line flag.

Some device libraries (in particular, Thrust) require this compliant preprocess to be used for compatibility. To support that, Viskores now adds the `/Zc:preprocessor` flag to compilations of device code. This is done by adding the compile option to the `viskores_exec` interface library.

Fixes: #276

### DIFF
--- a/CMake/ViskoresCompilerFlags.cmake
+++ b/CMake/ViskoresCompilerFlags.cmake
@@ -81,6 +81,14 @@ if(VISKORES_COMPILER_IS_MSVC)
   endif()
 endif()
 
+# Use a standard-compliant preprocessor (required for recent versions of Thrust).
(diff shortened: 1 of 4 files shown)


vicentebolea (approved):
This was interesting to learn

PR #281, opened 2026-04-02
Fix vulnerabilities reported by OpenSSF

kmorel:
https://github.com/Viskores/viskores/security/code-scanning/35


vicentebolea (approved):
Looks good, in the future dependabot should catch these

PR #282, opened 2026-04-07
Build CI in /tmp directory

kmorel:
We have been having some issues with object files disappearing, and this may be an issue with a networked file system. To get around this issue, compile to the /tmp directory instead.

### DIFF
--- a/.gitlab/ci/config/gitlab_ci_setup.cmake
+++ b/.gitlab/ci/config/gitlab_ci_setup.cmake
@@ -25,7 +25,11 @@ endif ()
 
 # Set up the source and build paths.
 set(CTEST_SOURCE_DIRECTORY "$ENV{CI_PROJECT_DIR}")
-set(CTEST_BINARY_DIRECTORY "${CTEST_SOURCE_DIRECTORY}/build")
+if (DEFINED ENV{CTEST_BINARY_DIRECTORY})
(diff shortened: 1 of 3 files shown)


kmorel:
@vicentebolea This change should fix the intermittent errors we have been seeing on ubuntu2204_hip_kokkos43. Please review.

.gitlab/ci/config/gitlab_ci_setup.cmake
```diff
@@ -25,7 +25,11 @@ endif ()
 
 # Set up the source and build paths.
 set(CTEST_SOURCE_DIRECTORY "$ENV{CI_PROJECT_DIR}")
-set(CTEST_BINARY_DIRECTORY "${CTEST_SOURCE_DIRECTORY}/build")
+if (BUILD_IN_TMP)
+  set(CTEST_BINARY_DIRECTORY "/tmp/build")
+else ()
+  set(CTEST_BINARY_DIRECTORY "${CTEST_SOURCE_DIRECTORY}/build")
+endif ()
```
vicentebolea, line 32: perhaps we can check if a CTEST_BINARY_DIRECTORY env variable is set and use its value for its corresponding cmake variable.

kmorel: That makes sense. I adjusted the scripts to do that.

vicentebolea (approved):

PR #283, opened 2026-04-09
annotation occlusion

nicolemarsaglia:
fix pseudocolor annotation depth against translucent volume rendering

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,9 +20,9 @@
 
 .docker_image: &docker_image
   variables:
-    CCACHE_BASEDIR: "${CI_PROJECT_DIR}"
+    CCACHE_BASEDIR: "/tmp"
     CCACHE_COMPILERCHECK: "content"
-    CCACHE_DIR: "$CI_PROJECT_DIR/.ccache"
+    CCACHE_DIR: "/tmp/.ccache"
     CCACHE_HASHDIR: "true"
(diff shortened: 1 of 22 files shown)


kmorel:
Thanks for taking on this fix, @nicolemarsaglia. This is a non-trivial fix.

It looks like the fix involves adding a secondary depth buffer that is used to determine whether to blend a fragment in front of or behind what is in the color buffer. Do we need this as a second depth buffer? Couldn't we just maintain the one depth buffer and do the behind blending if the color buffer is transparent or cull if it is opaque (which is what would happen with color blending anyway)?

nicolemarsaglia:
Mmmm ya, I wasn't fully understanding the code and that max distance is our opaque geometry. I definitely made it more complicated than it needed to be. 

I'm gonna take another swing at it. Thanks!

nicolemarsaglia:
@kmorel Made a more straightforward change where the volume writes to the canvas if the depth is closer than what is currently there. 

Let me know if you see any issues.

kmorel (approved):
That simplified a lot more than I was expecting. Glad it is working.

It would behoove you to add a test to make sure we don't break Ascent's functionality in the future. But other than that, this looks good.

kmorel:
@nicolemarsaglia, before merging, can you squash the commits together. (Instructions available at https://github.com/Viskores/viskores/blob/main/CONTRIBUTING.md#squash-commits.)

Issue #219, opened 2025-11-15
CUDA can't be enabled for current MSVC versions


Flashtim:
I have been trying to build Viskores on a couple of different Windows 11 machines using MSVC with the Viskores_ENABLE_CUDA option set to true.
While I can generate project files for Visual Studio 2022 with no errors or warnings in Cmake, Viskores fails to be compiled (2693 errors) due to issues with CUDA/Thrust.

These errors mostly repeat for a whole lot of files:
- ""thrust::detail" is ambiguous"
- "a template argument list is not allowed in a declaration of a primary template"
- "class or struct definition is missing"
- "make_projection_symbol: no fundamental base class"
- "'cudafe++' died with status 0xC0000409"

Further, there are issues with CUDA targets, e.g:

[...]

PR #290, opened 2026-04-17
Fix issues when compiling with Cuda 13.2

kmorel:
Header files for thrust components like distance and pair need to be explicitly included.

This should help with #219 if not fix it entirely.


kmorel:
@spyridon97 Please review these VIskores changes that were required to compile on the nvidia/cuda:13.2.0-devel-ubuntu22.04 image.

cc @vicentebolea. At some point we should probably have an updated dashboard.

spyridon97:
+2

spyridon97 (approved):
+2 Assuming that you have checked it locally.

PR #292, opened 2026-04-21
Improve BOV reader format support and tests

dpugmire:
Expand BOV parsing to handle scalar/vector fields, endian settings, byte offsets, absolute data paths, and zonal centering. Add validation for malformed or unsupported headers and reject divided bricks explicitly.

Move path absolute checks into FileUtils, add MakeAbsolutePath, and cover both helpers in FileUtils tests. Extend BOV reader tests with generated fixtures for formats, vectors, endian/offset handling, zonal data, absolute paths, and error cases.

### DIFF
--- a/viskores/io/BOVDataSetReader.cxx
+++ b/viskores/io/BOVDataSetReader.cxx
@@ -18,74 +18,644 @@
 
 #include <viskores/io/BOVDataSetReader.h>
 
+#include <viskores/cont/ArrayHandleRuntimeVec.h>
 #include <viskores/cont/DataSetBuilderUniform.h>
 #include <viskores/io/ErrorIO.h>
+#include <viskores/io/FileUtils.h>
+#include <viskores/io/internal/Endian.h>
 
+#include <algorithm>
+#include <cctype>
 #include <fstream>
+#include <limits>
 #include <sstream>
+#include <vector>
 
 namespace
 {
 
 enum class DataFormat
 {
-  ByteData,
-  ShortData,
-  IntegerData,
-  FloatData,
-  DoubleData
+  Byte,
+  Short,
+  Integer,
+  Float,
+  Double
 };
 
-template <typename T>
-void ReadBuffer(const std::string& fName, const viskores::Id& sz, std::vector<T>& buff)
+enum class Centering
 {
-  FILE* fp = fopen(fName.c_str(), "rb");
-  size_t readSize = static_cast<size_t>(sz);
-  if (fp == nullptr)
+  Nodal,
+  Zonal
+};
+
+struct BOVHeader
+{
+  BOVHeader()
+    : DataSize(0, 0, 0)
+    , Origin(0, 0, 0)
+    , BrickSize(1, 1, 1)
   {
+  }
+
+  std::string DataFile;
+  std::string VariableName;
+  viskores::Id3 DataSize;
+  viskores::Vec3f Origin;
+  viskores::Vec3f BrickSize;
+  viskores::IdComponent NumComponents = 1;
+  viskores::Id ByteOffset = 0;
+  DataFormat Format = DataFormat::Float;
+  Centering FieldCentering = Centering::Nodal;
+  bool HasDataFile = false;
+  bool HasDataSize = false;
+  bool HasFormat = false;
+  bool HasVariable = false;
+  bool HasBrickSize = false;
+  bool HasEndian = false;
+  bool DataIsLittleEndian = false;
+};
+
+std::string Trim(const std::string& value)
+{
+  auto begin = std::find_if_not(
+    value.begin(), value.end(), [](unsigned char ch) { return std::isspace(ch) != 0; });
+  auto end = std::find_if_not(
+    value.rbegin(), value.rend(), [](unsigned char ch) { return std::isspace(ch) != 0; });
+
+  if (begin >= end.base())
+    return std::string();
+  return std::string(begin, end.base());
+}
+
+std::string ToUpper(const std::string& value)
+{
+  std::string result = value;
+  std::transform(result.begin(),
+                 result.end(),
+                 result.begin(),
+                 [](unsigned char ch) { return static_cast<char>(std::toupper(ch)); });
+  return result;
+}
+
+std::string NormalizeKey(const std::string& key)
+{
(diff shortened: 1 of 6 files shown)


kmorel:
I'm listed as a reviewer for this PR and the tests pass, but the PR is still in draft mode. Take it out of draft and drop a comment when you are ready for a review.

dpugmire:
@kmorel 
I think is ready now for a review.

viskores/io/testing/UnitTestBOVDataSetReader.cxx
```diff
@@ -45,33 +51,323 @@ inline viskores::cont::DataSet readBOVDataSet(const char* fname)
...
+  VISKORES_TEST_ASSERT(field.GetNumberOfValues() == static_cast<viskores::Id>(expected.size()),
+                       "Incorrect vector field value count");
+
+  auto array =
+    field.GetData().template AsArrayHandle<viskores::cont::ArrayHandle<viskores::Vec<T, 3>>>();
+  auto portal = array.ReadPortal();
+  for (viskores::Id i = 0; i < static_cast<viskores::Id>(expected.size()); ++i)
+    VISKORES_TEST_ASSERT(test_equal(portal.Get(i), expected[static_cast<std::size_t>(i)]),
+                         "Incorrect vector field value");
+}
```
kmorel, line 131: This is all fine, but it would be much easier to test the field by wrapping the `std::vector` in an `ArrayHandle` and comparing the array handles.


```suggestion
template <typename T>
void CheckField(const viskores::cont::DataSet& ds,
                const std::string& fieldName,
                viskores::cont::Field::Association association,
                const std::vector<T>& expected)
{
  const auto& field = ds.GetField(fieldName, association);
  auto expectedAH = viskores::cont::make_ArrayHandle(expected, viskores::CopyFlag::Off);
  VISKORES_TEST_ASSERT(test_equal_ArrayHandles(field.GetData(), expectedAH),
                       "Incorrect scalar value for field ", "fieldName");
}
```

viskores/io/testing/UnitTestFileUtils.cxx
```diff
@@ -153,6 +153,39 @@ void TestMergePaths()
...
+  VISKORES_TEST_ASSERT(!IsAbsolutePath("relative/path"), "Should reject relative path");
+  VISKORES_TEST_ASSERT(!IsAbsolutePath(""), "Should reject empty path");
+#ifdef _MSC_VER
+  VISKORES_TEST_ASSERT(IsAbsolutePath("C:\\windows\\path"),
+                       "Should detect Windows drive absolute path");
+  VISKORES_TEST_ASSERT(IsAbsolutePath("\\windows\\path"),
+                       "Should detect Windows rooted absolute path");
+  VISKORES_TEST_ASSERT(!IsAbsolutePath("windows\\path"), "Should reject Windows relative path");
+  VISKORES_TEST_ASSERT(!IsAbsolutePath("C:windows\\path"),
+                       "Should reject Windows drive-relative path");
```
kmorel, line 174: Windows should probably check both forward and backward slashes.

viskores/io/FileUtils.cxx
```diff
@@ -125,6 +127,35 @@ bool CreateDirectoriesFromFilePath(const std::string& filePath)
...
+{
+  if (filePath.empty())
+    return false;
+#ifdef _WIN32
+  if (filePath.size() > 2 && filePath[1] == ':' && (filePath[2] == '/' || filePath[2] == '\\'))
+    return true;
+  return (filePath[0] == '/' || filePath[0] == '\\');
+#else
+  return filePath[0] == '/';
+#endif
```
kmorel, line 146: C++17 has a built in mechanism to check for absolute path.

```suggestion
  std::filesystem::path path(filePath);
  return path.is_absolute();
```

That said, per [the documentation] on windows paths that start with `/` are not absolute.

[the documentation]: https://en.cppreference.com/cpp/filesystem/path/is_absrel

viskores/io/FileUtils.cxx
```diff
@@ -125,6 +127,35 @@ bool CreateDirectoriesFromFilePath(const std::string& filePath)
...
+
+  char currentDirectory[4096];
+#ifdef _WIN32
+  if (_getcwd(currentDirectory, sizeof(currentDirectory)) == nullptr)
+#else
+  if (getcwd(currentDirectory, sizeof(currentDirectory)) == nullptr)
+#endif
+    throw viskores::cont::ErrorBadValue("Unable to get current working directory");
+
+  return MergePaths(currentDirectory, filePath);
```
kmorel: C++17 has a [built-in mechanism to make a path absolute].

```suggestion
  return std::filesystem::absolute(filePath);
```

[built-in mechanism to make a path absolute]: https://en.cppreference.com/cpp/filesystem/absolute

viskores/io/BOVDataSetReader.cxx
```diff
@@ -20,62 +20,394 @@
...
+  if (!header.HasDataFile || header.DataFile.empty())
+    throw viskores::io::ErrorIO("Missing required DATA_FILE option");
+  if (!header.HasDataSize)
+    throw viskores::io::ErrorIO("Missing required DATA_SIZE option");
+  if (!header.HasFormat)
+    throw viskores::io::ErrorIO("Missing required DATA_FORMAT option");
+  if (!header.HasVariable || header.VariableName.empty())
+    throw viskores::io::ErrorIO("Missing required VARIABLE option");
+  if (header.NumComponents != 1 && header.NumComponents != 3)
+    throw viskores::io::ErrorIO("Unsupported DATA_COMPONENTS value");
```
kmorel: Is the limitation of supporting only 1 or 3 components true for the BOV format or only for Viskores? If BOV is supposed to support any number of components, then you can use an `ArrayHandleRuntimeVec` to specify the number of components in an `ArrayHandle`. If you construct an `ArrayHandleRuntimeVec` and put it in a `Field`, then it can be used interchangeably with a basic `ArrayHandle` of the appropriate scalar or vec length in a later filter.

See https://viskores.readthedocs.io/en/latest/memory-layout.html#runtime-vec-arrays for the documentation.

kmorel (approved):
Looks good. My comments are more suggestions than anything.

PR #294, opened 2026-04-21
Fix issues when compiling with Cuda 13.2 (backport version 1.0)

kmorel:
Header files for thrust components like distance and pair need to be explicitly included.


vicentebolea (approved):
Error in macos build is independent of this change an CI spicific

PR #295, opened 2026-04-22
Add initial Python bindings

ghweber:
### Summary

This adds an initial `nanobind`-based Python binding layer for Viskores. The
bindings are organized under the existing Viskores namespaces and expose the
core classes needed to create Viskores data sets in Python, run Viskores
algorithms and filters, and inspect the results from Python.

The implementation includes:

- A `python_bindings/` build and packaging entry point.
- Bindings for core `viskores`, `viskores::cont`, `viskores::filter`,
  `viskores::io`, `viskores::rendering`, and `viskores::source` functionality.
- NumPy interoperability for supported scalar, vector, coordinate, and field
  array value types through `nanobind` ndarrays.
- Python tests registered with CTest when Python bindings and testing are
  enabled.
- Example scripts demonstrating data-set construction, filters, rendering, and
  selected analysis workflows.
- Source and wheel packaging support for building the Python bindings against a
  Viskores installation.

#### Binder

Binder was explored because it could, in principle, track changes in the
underlying Viskores classes with less manual rebinding work.

In practice, Binder still appeared to need a curated adapter layer between the
raw Viskores API and the desired Python API. Once that adapter layer is needed,
much of the automatic class-surface tracking advantage is lost because the
adapter layer must still be maintained as the C++ API changes.

For this branch, that made Binder less attractive than a direct binding layer
that intentionally exposes the selected Viskores API surface.

#### cppyy

`cppyy` remained interesting because of its low-maintenance, header-driven
model.

In practice, the Apple Silicon/macOS Cling/rootcling toolchain problems were
severe enough that it never reached a fair Viskores-specific evaluation. The
packaged `cppyy` stack failed before Viskores-specific testing, and a
source-built `cppyy-cling` stack still failed in `rootcling_stage1` dictionary
generation on a non-Viskores test case.

That made `cppyy` a poor choice for the current bindings effort, though it may
still be worth revisiting if the platform/toolchain situation improves.

### Why nanobind

`nanobind` gives direct control over the Python surface while keeping the
implementation relatively compact. It supports a public API that tracks
Viskores C++ classes and namespaces closely, avoids the adapter-heavy Binder
direction, and avoids the toolchain blockers seen in the `cppyy` exploration.

It was the most practical path to a real, reviewable binding surface for this
initial Python support.

### Design decisions

- The public Python API should prefer real Viskores classes and C++-style names.
- Python-only convenience surfaces should complement, not replace, the main
  Viskores API.
- Example-specific workflows should stay in examples, not in the public
  bindings.
- The active binding/runtime path is pure `nanobind`.
- Array conversion uses `nanobind` ndarrays rather than the direct NumPy C API.
- The default extension build uses the full Python ABI so the repository can
  keep its CMake 3.15 baseline.
- The `nanobind` stable-ABI (`abi3`) path remains available explicitly via
  `VISKORES_PYTHON_STABLE_ABI=ON`; that mode currently requires Python 3.12 or
  newer and CMake FindPython support for `Python::SABIModule`.

### Validation status

- Built the in-tree bindings:
  - `cmake --build /Users/ghweber/devel/viskores-build --target viskores_python_bindings --parallel 4`
- Ran the Python binding CTest subset:
  - `ctest --test-dir /Users/ghweber/devel/viskores-build -R '^(PythonBindingsSmokeTest|(UnitTestPythonWrapper|PythonWrapper).*)$' --output-on-failure`
  - `106/106` Python wrapper tests passed.
- Built and smoke-tested an opt-in stable-ABI wheel:
  - `viskores-1.1.9999-cp312-abi3-macosx_26_0_arm64.whl`
- Ran the license check after adding `python_bindings/version.txt` as a
  data-file exception:
  - `cmake -DViskores_SOURCE_DIR=/Users/ghweber/devel/viskores -P /Users/ghweber/devel/viskores/CMake/ViskoresCheckLicense.cmake`

### Deferred items

- `BufferState` still uses an interim holder implementation in the bindings and
  should receive a deeper design review later.
- `UnitTestMultiMapper.cxx` remains deferred pending a decision on exposing
  low-level mapper `RenderCells(...)`.
- Making `abi3` the default would require newer CMake FindPython support for
  `Python::SABIModule`. Once this PR is ready to move from draft to full
  consideration, create and link a GitHub issue to discuss whether raising the
  minimum CMake version for default `abi3` support is acceptable.


CMakeLists.txt
```diff
@@ -565,6 +566,10 @@ endif()
 # Build examples
 add_subdirectory(examples)
 
+if (Viskores_ENABLE_PYTHON_BINDINGS)
+  add_subdirectory(python_bindings)
+endif()
```
kmorel: Subdirectories should use the Viskores module facility rather than directly using `add_subdirectory`. To do that, add a file named `viskores.module` to the `python_bindings` directory. The Viskores build will automatically pick that up and add it to the build. Documentation of modules is in [docs/Modules.md](https://github.com/Viskores/viskores/blob/main/docs/Modules.md). The contents of the `viskores.module` file would probably be something like this:

```
NAME
  viskores_python_bindings
GROUPS
  PythonBindings
PRIVATE_DEPENDS
  viskores_cont
  viskores_source
  viskores_filter_core
  viskores_filter_contour
  viskores_filter_field_conversion
  viskores_filter_scalar_topology
  viskores_filter_vector_analysis
  viskores_rendering
```

To link the building of the module to the `Viskores_ENABLE_PYTHON_BINDINGS`, you can use the `viskores_module_force_group` CMake function to enable/disable the `PythonBindings` group (in the `viskores.module` example I have above).

```cmake
viskores_module_force_group(PythonBondings ENABLE_OPTION Viskores_ENABLE_PYTHON_BINDINGS)
```

python_bindings/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+  RESULT_VARIABLE Viskores_PYTHON_NUMPY_RESULT
+)
+if (NOT Viskores_PYTHON_NUMPY_RESULT EQUAL 0)
+  message(FATAL_ERROR "NumPy is required to build Viskores Python bindings.")
+endif()
+
+set(Viskores_PYTHON_PACKAGE_DIR "${Viskores_BINARY_DIR}/python_bindings/viskores")
+file(MAKE_DIRECTORY "${Viskores_PYTHON_PACKAGE_DIR}")
+
+add_library(_viskores MODULE viskores_module.cxx)
```
kmorel: `viskores_module.cxx` may be confused with the CMake Viskores modules. Maybe name this `viskores_python_bindings.cxx`?

python_bindings/wheel/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
```
kmorel, line 11: This file is causing [one of the tests] to fail. It has to start with the following text:

```cmake
##=============================================================================
##
##  The contents of this file are covered by the Viskores license. See
##  LICENSE.txt for details.
##
##  By contributing to this file, all contributors agree to the Developer
##
##=============================================================================
```

[one of the tests]: https://open.cdash.org/tests/2518389501

python_bindings/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+
+##============================================================================
+##  Copyright (c) Kitware, Inc.
+##  All rights reserved.
+##  See LICENSE.txt for details.
+##
+##  This software is distributed WITHOUT ANY WARRANTY; without even
+##  the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+##  PURPOSE.  See the above copyright notice for more information.
+##============================================================================
```
kmorel: This second license statement is not necessary. This is the old statement from the VTK-m days. It is fine to slap a secondary copyright statement after the main one at the top (if, for example, LBL wants one), but no need to perpetuate this old one. We just didn't delete them from the old files because we are lazy.

python_bindings/viskores_module.cxx
```diff
@@ -0,0 +1,2010 @@
...
+#include <viskores/filter/scalar_topology/worklet/contourtree_augmented/ProcessContourTree.h>
+#include <viskores/rendering/Actor.h>
+#include <viskores/rendering/Camera.h>
+#include <viskores/rendering/CanvasRayTracer.h>
+#include <viskores/rendering/MapperRayTracer.h>
+#include <viskores/rendering/MapperVolume.h>
+#include <viskores/rendering/MapperWireframer.h>
+#include <viskores/rendering/Scene.h>
+#include <viskores/rendering/View3D.h>
+#include <viskores/source/Tangle.h>
```
kmorel: This is a big-ass file. How was it created?

As Viskores changes the bindings will have to change with them. Is there a way to, say, easily add a new object?

kmorel (changes requested):
This is cool. I have not done a full review, but Python bindings are something we have wanted for a while.

I haven't done a full review, but I have some comments and requested changes so far.

Also, please add a changelog file to `docs/changelog`.

python_bindings/viskores_module.cxx
ghweber: I at least changed it to have files per module. Still iterating, but hopefully better. Unfortunately my attempts to automate binding generation were not successful as Viskores seems a bit too complex for their automated handling of things.

ghweber:
Python stable-ABI note: the branch now keeps abi3 builds available behind VISKORES_PYTHON_STABLE_ABI=ON, but defaults to the full Python ABI so the repo can keep its CMake 3.15 baseline. Making abi3 the default would require newer CMake FindPython support for Python::SABIModule, so when this PR moves from draft to full consideration we should create/link a GitHub issue to discuss whether raising the minimum CMake version for that is acceptable.

ghweber:
Python binding update pushed in `2c1d21de1`:\n\n- Expanded NumPy/Viskores array interop for the NumPy dtypes that map cleanly to Viskores scalar and fixed-vector arrays: int8/uint8 through int64/uint64, plus float32/float64, for scalar and vec2/vec3/vec4 arrays.\n- Added `viskores.cont.array_from_numpy` / `asnumpy`, registered matching SOA and recombined array classes, and allowed registered Viskores arrays to flow through field and coordinate APIs.\n- Preserved float32/float64 coordinate arrays in explicit dataset construction and added focused CTest coverage for the new interop paths.\n- Verified the in-tree stable-ABI nanobind build and the wheel path with `VISKORES_PYTHON_STABLE_ABI=ON`; the local Python binding CTest subset passed 106/106.\n\nFollow-up for when this moves out of draft: we should create a GitHub issue to discuss raising the minimum CMake version once the project is ready. CMake 3.26+ would let us rely on FindPython's `Python::SABIModule` support for the abi3/stable-ABI path instead of keeping that capability behind an optional path constrained by the repo's current CMake 3.15 baseline.

vicentebolea:
@ghweber adios2 uses nanobind as well for its c++-> python bindings, https://github.com/ornladios/ADIOS2/tree/master/bindings/Python We do have a C++ special api to accomodate to nanobind/pybind.

ghweber:
Python binding packaging update pushed in `98ca0ae9b`:\n\n- Kept generated wheel output out of the source tree by defaulting `bdist_wheel` output to `../viskores-build/python_bindings/dist/`.\n- Moved setuptools build and egg metadata under `../viskores-build/python_bindings/setuptools/` so local wheel builds do not recreate `python_bindings/dist/` or source-tree metadata artifacts.\n- Updated the Python binding README examples to use the build-tree locations and document `VISKORES_PYTHON_BUILD_ROOT` / `bdist_wheel --dist-dir` overrides.\n- Updated my local `viskores-experiments` environment to consume the rebuilt wheel from the build-tree path.\n\nValidation for this head is green: GitHub Actions format/macOS/Windows, Doxygen user guide, Read the Docs, CDash, Frank CI, local `git diff --check`, local `python3 -m py_compile python_bindings/setup.py`, and the local Viskores license check all pass.

kmorel:
Um. I'm not sure what to do with this. Your LLM just barfed 20K lines of code. I can't seriously review all that. And even if I trusted the code now, I don't know how we could reasonably maintain this.

I expect there to be heavily customized code for some of the classes in the `viskores::cont` library concerning array handles and cell sets. Maybe that is 1K LOC (hopefully a lot less). But this is the reason there needs to be a deterministic way to build bindings from code in the filter and IO libraries.

ghweber:
@kmorel I am using this as a learning experience myself and am possibly getting a bit carried away. Sorry about that. Will see what I can do/possibly change. But this is a bit of an experiment whether I can get LLMs to do what I want. Thanks for the comment and let me see how this goes.

Feel free to ignore this pull request at the moment as it is just not ready yet. On the other hand, your comments like the one above help me avoid going down rabbit holes even further than I am already going.

ghweber:
@kmorel please note also that in addition to the bindings being verbose, I am creating these bindings by replicating most Viskores tests (tests for IO, filters, rendering, skipping internals) in Python, so that also adds bulk to the PR.

kmorel:
> @kmorel please note also that in addition to the bindings being verbose, I am creating these bindings by replicating most Viskores tests (tests for IO, filters, rendering, skipping internals) in Python, so that also adds bulk to the PR.

Fair point. That's a good chunk of the PR and easier to maintain even if it is excessive. Still, we need a mechanism to keep the bindings up to date.

ghweber:
Update after the latest Python binding cleanup pass:

- Pushed commit `1aeeaba05` (`Automate Python binding manifest coverage`).
- The binding manifest now tracks all scanned filter/IO candidates: `119` generated, `23` manual, `41` excluded, `0` missing.
- The generator now supports explicit constructors, instance methods, static methods, scoped/top-level enums, Python base classes, registration groups, and reusable argument/return converters.
- Several classes that were previously manual are now generated, including `Field`, `CoordinateSystem`, `DataSet`, `PartitionedDataSet`, `ColorTable`, `Probe`, `ContourTreeUniformDistributed`, `DataSetBuilderUniform`, and `DataSetBuilderExplicitIterative`.
- Python-only helpers were moved out of the C++-named API into `viskores.python_convenience` where possible, and the current known C++/Python API deviations are documented in `python_bindings/cpp_api_deviations.md`.

Local verification:

- `cmake --build /Users/ghweber/devel/viskores-build`
- `ctest --test-dir /Users/ghweber/devel/viskores-build -R UnitTestPythonWrapper --output-on-failure` (`98/98` passed)
- `PythonBindingManifestAudit` passed as part of the focused checks.

At this point the remaining work is largely design/policy rather than mechanical conversion. I would especially appreciate input from other Viskores developers on these points:

1. Whether the manifest/manual/excluded ledger is the right long-term way to track Python binding coverage.
2. How closely Python should model templated holders such as `UnknownArrayHandle`, `UncertainArrayHandle`, and possibly `ArrayHandle`, versus using NumPy as the Python-facing storage creation path.
3. Whether exact-length Python tuples are acceptable for tuple-like C++ value types such as `Id2`, `Id3`, and `Vec3f` in this first iteration.
4. Whether abstract API bases such as `ImageWriterBase` should stay exposed for C++ API fidelity, even if they remain manual.
5. Whether flow filters should remain excluded until there is a broader Python design for seeds, vector fields, and integration controls.
6. Whether the current ownership wrappers for rendering classes are acceptable for this initial PR, or whether we should narrow rendering scope before merge.

python_bindings/binding_manifest/cont.yml
```diff
@@ -0,0 +1,245 @@
...
+  constructors:
+    - ()
+    - (src: const viskores::cont::UnknownArrayHandle&)
+  registrationGroup: early
+
+- class: viskores::cont::ArrayHandleSOA<viskores::Vec<T, N>>
+  doc: Viskores ArrayHandleSOA wrapper.
+  header: viskores/cont/ArrayHandleSOA.h
+  binding: manual
+  reason: Registered as selected scalar/vector template instantiations with NumPy helpers while the public ArrayHandle template model is still open.
```
kmorel, line 29: I'm surprised to see `UncertainArrayHandle` or any of the `ArrayHandle*` classes with bindings at all. How do bindings handle templates?

I don't expect to need these templated array classes available from Python. Just have a special implementation of `UnknownArrayHandle` that goes to and from numpy arrays (which I think you already have).

python_bindings/binding_manifest/cont.yml
```diff
@@ -0,0 +1,245 @@
...
+##
+##  By contributing to this file, all contributors agree to the Developer
+##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+##
+##=============================================================================
+
+- class: viskores::cont::UnknownArrayHandle
+  doc: Type-erased Viskores array handle.
+  binding: manual
+  reason: Requires NumPy conversion, runtime dispatch over registered array/storage families, and the current Unknown/Uncertain array-handle design.
```
kmorel, line 14: What is the point of having manual bindings in here at all? It seems like the parser is just going to ignore these.

kmorel (commented):
I'm warming up to, but not convinced about, the yaml descriptions. We may want to think about tweaks to the header files to make it easy to pull out parts to bind.

But more importantly, this PR is still way too overwhelming. I need baby steps to really think about and discuss the bindings.

Where we need to start is a "simple" PR that wraps `viskores::cont::UnknownArrayHandle` and provides a single Python test that pushes a numpy array into an `UnknownArrayHandle` and then gets a numpy array out of the `UnknownArrayHandle`. That is a core functionality that has to be right for most other things to work at all, and there will likely be a lot of discussion on just that.

ghweber:
Thanks, this makes sense. I agree this PR is still too large to review productively as a merge candidate, even though it has been useful as an exploratory branch.

I like the `UnknownArrayHandle` + NumPy round-trip as the core starting point. I propose opening a new, much smaller PR for that rather than trying to reshape this PR in place. That new PR should focus on the core array path, but include just enough surrounding context to verify it in a real Viskores workflow:

1. Wrap `viskores::cont::UnknownArrayHandle` with NumPy input/output.
2. Add a minimal `DataSet` path that attaches a NumPy-backed field.
3. Run one simple filter, perhaps `Contour` for a basic isosurface extraction or another small filter, so we can see data move Python NumPy -> Viskores field/data set -> filter -> Python-visible result.

That should keep the review focused on the key ownership/type/mutability questions around arrays while still testing the core use case rather than only an isolated array object.

On the manifest: the intent was not that `manual` or `excluded` entries are generated. The reason they are listed is to support a coverage audit. The current branch has a `PythonBindingManifestAudit` CTest test that runs the manifest tool with `--fail-on-missing`; if a scanned filter/IO class is not generated, manual, or excluded, the test fails. So the manual/excluded entries are meant as a ledger of explicit decisions, not generator input. That said, I agree this is probably too much for the first small PR. We can leave that machinery in this exploratory branch for later, or split generated binding descriptions from a separate coverage/status ledger if that is clearer.

I will treat this PR as a reference/exploration branch and start a new focused PR for the `UnknownArrayHandle`/NumPy/DataSet/simple-filter path.

python_bindings/binding_manifest/cont.yml
ghweber: I agree. The current branch exposed selected concrete template instantiations mostly as part of exploring the API surface, but that is probably not the right first public Python API. For the smaller new PR I would focus on a special `UnknownArrayHandle` binding that converts to/from NumPy arrays and defer `ArrayHandle*` / `UncertainArrayHandle` unless a clear Python use case requires them.

python_bindings/binding_manifest/cont.yml
ghweber: The `manual` entries are there for audit coverage, not generation. They let the manifest audit fail on truly missing scanned classes while accepting classes that are intentionally hand-bound or excluded. But I agree this is not needed in the first focused PR; it can stay in this exploratory branch until we decide whether the coverage ledger should live in the manifest or in a separate file.

kmorel:
> Thanks, this makes sense. I agree this PR is still too large to review productively as a merge candidate, even though it has been useful as an exploratory branch.
> 
> I like the `UnknownArrayHandle` + NumPy round-trip as the core starting point. I propose opening a new, much smaller PR for that rather than trying to reshape this PR in place.

I agree.

> That new PR should focus on the core array path, but include just enough surrounding context to verify it in a real Viskores workflow:
> 
> 1. Wrap `viskores::cont::UnknownArrayHandle` with NumPy input/output.
> 2. Add a minimal `DataSet` path that attaches a NumPy-backed field.
> 3. Run one simple filter, perhaps `Contour` for a basic isosurface extraction or another small filter, so we can see data move Python NumPy -> Viskores field/data set -> filter -> Python-visible result.

Each of these three steps need to be their own PR. They each represent something tricky that should have a good amount of scrutiny. For the first step, I really just want to see the numpy/ArrayHandle interface.

Somewhere between 1 and 3 we'll add one or more CI builds that test the Python bindings. (I might need @vicentebolea's help to get images with Python.)

ghweber:
@kmorel Agreed. We created #308 as the focused follow-up for this first step: `viskores::cont::UnknownArrayHandle` with NumPy input/output.

I am closing this broader exploratory PR because it is too large to review productively as a single merge candidate. We will bring the remaining pieces back as smaller PRs, split into separate steps such as DataSet/Field support, filters, and any generated-binding or manifest machinery once the core array interface has had enough review.

PR #301, opened 2026-04-27
Fix VTK reader edge cases in FIELDs, split cells, and structured grids

dpugmire:
Fix VTK reader bugs in FIELD parsing, split-cell data, and structured dims

Skip rejected FIELD array payloads so later arrays are parsed correctly.
When POLY_LINE, TRIANGLE_STRIP, or POLY_VERTEX cells are split into
multiple Viskores cells, permute full CELL_DATA tuples instead of
treating them as scalars. Accept degenerate structured dimensions on
any axis, not just X or XY layouts.

Add reader regressions for mismatched FIELD arrays, multi-component
CELL_DATA on split cells, and degenerate STRUCTURED_POINTS,
RECTILINEAR_GRID, and STRUCTURED_GRID inputs.

### DIFF
--- a/viskores/io/testing/UnitTestVTKDataSetReader.cxx
+++ b/viskores/io/testing/UnitTestVTKDataSetReader.cxx
@@ -16,13 +16,66 @@
 //  PURPOSE.  See the above copyright notice for more information.
 //============================================================================
 
+#include <viskores/cont/ArrayCopy.h>
+#include <viskores/cont/ArrayHandleRuntimeVec.h>
 #include <viskores/cont/testing/Testing.h>
+#include <viskores/io/FileUtils.h>
 #include <viskores/io/VTKDataSetReader.h>
 
+#include <array>
+#include <atomic>
+#include <cstdio>
+#include <cstdlib>
+#include <fstream>
 #include <string>
+
+#ifdef _MSC_VER
+#include <process.h>
+#else
+#include <unistd.h>
+#endif
 
 namespace
 {
+
+viskores::Id GetCurrentProcessId()
+{
+#ifdef _MSC_VER
+  return _getpid();
+#else
+  return getpid();
+#endif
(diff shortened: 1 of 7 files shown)


viskores/io/testing/UnitTestVTKDataSetReader.cxx
```diff
@@ -51,6 +88,99 @@ enum Format
...
+    VISKORES_TEST_ASSERT(out.good(), "Could not write ", this->FileName);
+  }
+
+  ~ScopedVTKTestFile() { std::remove(this->FileName.c_str()); }
+
+  const std::string& GetFileName() const { return this->FileName; }
+
+private:
+  std::string FileName;
+};
```
kmorel, line 128: Seems easier to just add files to the data submodule, but OK.

viskores/io/testing/UnitTestVTKDataSetReader.cxx
```diff
@@ -507,6 +637,170 @@ void TestSkppingStringFields(Format format)
...
+                            "0 0 0 1 0 0 2 0 0\n"
+                            "LINES 1 4\n"
+                            "3 0 1 2\n"
+                            "CELL_DATA 1\n"
+                            "VECTORS cellvec float\n"
+                            "1 2 3\n");
+
+  viskores::cont::DataSet ds = readVTKDataSet(vtkFile.GetFileName());
+
+  VISKORES_TEST_ASSERT(ds.GetNumberOfCells() == 2, "Polyline should expand into two cells");
```
kmorel: Viskores now supports poly line. It would be better to fix the reader to read polylines correctly.

(That said, triangle strips are still not supported, so that would be a good place to test permutation.)

kmorel (approved):
Looks good pending formatting fixes and CI.

PR #308, opened 2026-05-01
Add minimal Python UnknownArrayHandle bindings

ghweber:
## Summary

Adds an optional \`python_bindings\` module group (\`Viskores_ENABLE_PYTHON_BINDINGS=ON\`) that builds a [nanobind](https://github.com/wjakob/nanobind) extension with NumPy I/O for \`viskores::cont::UnknownArrayHandle\`:

- \`viskores.cont.array_from_numpy(array, allow_copy=False)\` — wrap a NumPy array as an \`UnknownArrayHandle\` sharing the same storage. Pass \`allow_copy=True\` to accept non-shareable inputs by copying.
- \`viskores.cont.asnumpy(array)\` / \`UnknownArrayHandle.asnumpy()\` — return a NumPy view of an \`UnknownArrayHandle\`. Zero-copy for basic and runtime-vec storage; element-by-element copy otherwise.

Design rationale is in \`python_bindings/src/cont_bindings.cxx\`.

## Naming Convention

CamelCase for C++-mirrored symbols (\`UnknownArrayHandle\`, \`GetNumberOfValues\`); lowercase for Python-specific helpers (\`array_from_numpy\`, \`asnumpy\`). Worth a sign-off in review.

## Testing

\`\`\`sh
cmake --build <build-dir> --target viskores_python_bindings -j 4
ctest --test-dir <build-dir> -R PythonBindingsUnknownArrayHandleNumPy --output-on-failure
\`\`\`

Stable-ABI variant: configure with \`-DVISKORES_PYTHON_STABLE_ABI=ON\` (requires Python >= 3.12 and CMake >= 3.26).

## Deferred

- **Install rule.** Build-tree only; packaging deferred until the binding surface grows.
- **CI hook.** No \`.gitlab-ci.yml\` job for \`Viskores_ENABLE_PYTHON_BINDINGS=ON\` yet.
- **Non-basic storage copy-path test.** \`asnumpy\` has two paths: zero-copy for \`ArrayHandleBasic\` and \`ArrayHandleRuntimeVec\` (covered by \`check_viskores_allocated_zero_copy\`), and an element-by-element fallback via \`CastAndCallForTypes\` for other storage layouts (\`ArrayHandleUniformPoints\`, \`ArrayHandleCartesianProduct\`, \`ArrayHandleSOAStride\`). The fallback compiles and the logic is straightforward, but exercising it at runtime requires constructing one of those non-basic arrays from Python, which is out of scope for this PR.

### DIFF
--- a/python_bindings/CMakeLists.txt
+++ b/python_bindings/CMakeLists.txt
@@ -1,0 +1,133 @@
+##============================================================================
+##  The contents of this file are covered by the Viskores license. See
+##  LICENSE.txt for details.
+##
+##  By contributing to this file, all contributors agree to the Developer
+##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+##============================================================================
+
+# Keep stable ABI support optional while the binding surface is small. This
+# lets developers check abi3 compatibility without making it the default build.
+option(VISKORES_PYTHON_STABLE_ABI "Build Python bindings with the CPython stable ABI" OFF)
+
+set(Viskores_PYTHON_COMPONENTS Interpreter Development)
+if (VISKORES_PYTHON_STABLE_ABI AND CMAKE_VERSION VERSION_GREATER_EQUAL 3.26)
+  list(APPEND Viskores_PYTHON_COMPONENTS Development.SABIModule)
+endif()
+find_package(Python 3.8 REQUIRED COMPONENTS ${Viskores_PYTHON_COMPONENTS})
+
+# Minimum nanobind version. nb::ndarray dtype dispatch and the stable-ABI
+# support used here require nanobind 2.0 or newer.
+set(Viskores_PYTHON_NANOBIND_MIN_VERSION "2.0.0")
+
+execute_process(
+  COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
+  OUTPUT_VARIABLE Viskores_PYTHON_NANOBIND_CMAKE_DIR
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  RESULT_VARIABLE Viskores_PYTHON_NANOBIND_RESULT
+)
+if (NOT Viskores_PYTHON_NANOBIND_RESULT EQUAL 0)
+  message(FATAL_ERROR
+    "nanobind ${Viskores_PYTHON_NANOBIND_MIN_VERSION} or newer is required to "
+    "build Viskores Python bindings. Install with: pip install 'nanobind>="
+    "${Viskores_PYTHON_NANOBIND_MIN_VERSION}'.")
+endif()
+
+find_package(nanobind ${Viskores_PYTHON_NANOBIND_MIN_VERSION} CONFIG REQUIRED
+  PATHS "${Viskores_PYTHON_NANOBIND_CMAKE_DIR}"
+  NO_DEFAULT_PATH
+)
+
+set(Viskores_PYTHON_NANOBIND_STABLE_ABI)
+if (VISKORES_PYTHON_STABLE_ABI)
+  if (Python_VERSION VERSION_LESS 3.12)
+    message(FATAL_ERROR
+      "VISKORES_PYTHON_STABLE_ABI requires Python 3.12 or newer.")
+  endif()
+  if (NOT TARGET Python::SABIModule)
+    message(FATAL_ERROR
+      "VISKORES_PYTHON_STABLE_ABI requires CMake FindPython support for "
+      "Python::SABIModule. Use CMake 3.26 or newer, or configure with "
+      "-DVISKORES_PYTHON_STABLE_ABI=OFF.")
+  endif()
+  set(Viskores_PYTHON_NANOBIND_STABLE_ABI STABLE_ABI)
+endif()
+
+# Minimum NumPy version. The binding uses the buffer protocol and the standard
+# ndarray flags (c_contiguous, aligned, writeable) that have been stable since
+# NumPy 1.20.
+set(Viskores_PYTHON_NUMPY_MIN_VERSION "1.20.0")
+
+# Pass the minimum version as an argv argument rather than substituting it into
+# the Python source string, so an unexpected character in the version literal
+# cannot interact with Python parsing. NumPy ships its own version comparator
+# (numpy.lib.NumpyVersion), so we don't need the third-party `packaging` module.
+execute_process(
+  COMMAND "${Python_EXECUTABLE}" -c
+    "import sys; from numpy.lib import NumpyVersion; import numpy; \
+sys.exit(0 if NumpyVersion(numpy.__version__) >= NumpyVersion(sys.argv[1]) else 1)"
+    "${Viskores_PYTHON_NUMPY_MIN_VERSION}"
+  RESULT_VARIABLE Viskores_PYTHON_NUMPY_RESULT
+)
+if (NOT Viskores_PYTHON_NUMPY_RESULT EQUAL 0)
+  message(FATAL_ERROR
+    "NumPy ${Viskores_PYTHON_NUMPY_MIN_VERSION} or newer is required to build "
+    "Viskores Python bindings.")
+endif()
+
+set(Viskores_PYTHON_PACKAGE_ROOT "${Viskores_BINARY_DIR}/python_bindings")
+set(Viskores_PYTHON_PACKAGE_DIR "${Viskores_PYTHON_PACKAGE_ROOT}/viskores")
+file(MAKE_DIRECTORY "${Viskores_PYTHON_PACKAGE_DIR}")
+
+# List Python source files explicitly. The set is small and listing files keeps
+# the dependency graph deterministic; revisit if the package grows.
+set(Viskores_PYTHON_PACKAGE_FILES
+  "${CMAKE_CURRENT_SOURCE_DIR}/viskores/__init__.py"
+  "${CMAKE_CURRENT_SOURCE_DIR}/viskores/cont/__init__.py"
+)
+set(Viskores_PYTHON_PACKAGE_STAMP
+  "${Viskores_BINARY_DIR}/python_bindings/.python_package_sync.stamp")
+
+add_custom_command(
+  OUTPUT "${Viskores_PYTHON_PACKAGE_STAMP}"
+  COMMAND "${CMAKE_COMMAND}" -E make_directory "${Viskores_PYTHON_PACKAGE_ROOT}"
+  COMMAND "${CMAKE_COMMAND}" -E copy_directory
+          "${CMAKE_CURRENT_SOURCE_DIR}/viskores"
+          "${Viskores_PYTHON_PACKAGE_DIR}"
+  COMMAND "${CMAKE_COMMAND}" -E touch "${Viskores_PYTHON_PACKAGE_STAMP}"
+  DEPENDS ${Viskores_PYTHON_PACKAGE_FILES}
+)
+add_custom_target(viskores_python_package_sync
+  DEPENDS "${Viskores_PYTHON_PACKAGE_STAMP}")
+
+nanobind_add_module(viskores_python_bindings ${Viskores_PYTHON_NANOBIND_STABLE_ABI} NOMINSIZE
+  src/cont_bindings.cxx
+  src/viskores_python_bindings.cxx
+)
+target_compile_features(viskores_python_bindings PRIVATE cxx_std_17)
+target_compile_definitions(viskores_python_bindings PRIVATE PY_SSIZE_T_CLEAN)
+target_include_directories(viskores_python_bindings
+  PRIVATE
+    "${CMAKE_CURRENT_SOURCE_DIR}/src"
+)
+target_link_libraries(viskores_python_bindings PRIVATE
+  viskores_cont
+  viskores_interop_python
+)
+add_dependencies(viskores_python_bindings viskores_python_package_sync)
+set_target_properties(viskores_python_bindings PROPERTIES
+  PREFIX ""
+  OUTPUT_NAME "_viskores"
+  LIBRARY_OUTPUT_DIRECTORY "${Viskores_PYTHON_PACKAGE_DIR}"
+  RUNTIME_OUTPUT_DIRECTORY "${Viskores_PYTHON_PACKAGE_DIR}"
+)
+
+set(_viskores_python_site_dir
+  "lib/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages")
+install(TARGETS viskores_python_bindings
+  LIBRARY DESTINATION "${_viskores_python_site_dir}/viskores"
+)
+install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/viskores/"
+  DESTINATION "${_viskores_python_site_dir}/viskores"
+  FILES_MATCHING PATTERN "*.py"
+)
--- a/python_bindings/examples/roundtrip_allow_copy.py
+++ b/python_bindings/examples/roundtrip_allow_copy.py
@@ -1,0 +1,40 @@
+##============================================================================
+##  The contents of this file are covered by the Viskores license. See
+##  LICENSE.txt for details.
+##
+##  By contributing to this file, all contributors agree to the Developer
+##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+##============================================================================
+
+import numpy as np
+
+import viskores
+import viskores.cont
+
+
+# Demonstrate the allow_copy=True kwarg on viskores.cont.array_from_numpy.
+# By default the binding requires a directly-shareable input (C-contiguous,
+# aligned, writable) so it can wrap the NumPy buffer with zero copy. Inputs
+# that don't meet those requirements are rejected with a descriptive error.
+# Pass allow_copy=True to let the binding make a contiguous, writable copy
+# automatically when the input is not directly shareable.
+
+base = np.arange(12, dtype=np.float32)
+
+# A non-contiguous slice would normally be rejected.
+sliced = base[::2]
+try:
+    viskores.cont.array_from_numpy(sliced)
+except RuntimeError as error:
+    print(f"non-contiguous default: rejected ({str(error).splitlines()[0]})")
+
+# allow_copy=True accepts the slice by copying it.
+unknown = viskores.cont.array_from_numpy(sliced, allow_copy=True)
+print(f"non-contiguous + allow_copy: {unknown}")
+np.testing.assert_array_equal(unknown.asnumpy(), sliced)
+
+# A directly-shareable input still takes the zero-copy path even with
+# allow_copy=True. Only inputs that need a copy are copied.
+unknown = viskores.cont.array_from_numpy(base, allow_copy=True)
+print(f"contiguous + allow_copy: shares storage = "
+      f"{np.shares_memory(unknown.asnumpy(), base)}")
--- a/python_bindings/src/cont_bindings.cxx
+++ b/python_bindings/src/cont_bindings.cxx
@@ -1,0 +1,77 @@
+//============================================================================
+//  The contents of this file are covered by the Viskores license. See
+//  LICENSE.txt for details.
+//
+//  By contributing to this file, all contributors agree to the Developer
+//  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+//============================================================================
+
+#include "viskores_python_bindings.h"
+
+#include <viskores/interop/python/ArrayHandleToNumPy.h>
+#include <viskores/interop/python/NumPyToArrayHandle.h>
+
+#include <viskores/cont/UnknownArrayHandle.h>
+
+#include <string>
+
+namespace vip = viskores::interop::python;
+
+namespace viskores::python::bindings
+{
+namespace
+{
+
+std::string UnknownArrayHandleRepr(const viskores::cont::UnknownArrayHandle& self)
+{
+  return "viskores.cont.UnknownArrayHandle(values=" +
+    std::to_string(self.GetNumberOfValues()) +
+    ", components=" + std::to_string(self.GetNumberOfComponentsFlat()) + ")";
+}
+
+} // namespace
+
+void BindCont(nb::module_& m)
+{
+  nb::class_<viskores::cont::UnknownArrayHandle>(m, "UnknownArrayHandle")
+    .def(nb::init<>())
+    .def("__len__", &viskores::cont::UnknownArrayHandle::GetNumberOfValues)
+    .def("__repr__", &UnknownArrayHandleRepr)
+    .def("GetNumberOfValues",
+         &viskores::cont::UnknownArrayHandle::GetNumberOfValues,
+         "Number of elements (tuples) in the array.")
+    .def("GetNumberOfComponentsFlat",
+         &viskores::cont::UnknownArrayHandle::GetNumberOfComponentsFlat,
+         "Number of scalar components per element, counting nested Vec components.")
+    .def("NewInstanceBasic",
+         &viskores::cont::UnknownArrayHandle::NewInstanceBasic,
+         "Allocate a new Viskores-managed array of the same value type.")
+    .def("DeepCopyFrom",
+         [](viskores::cont::UnknownArrayHandle& self,
+            const viskores::cont::UnknownArrayHandle& source) { self.DeepCopyFrom(source); },
+         nb::arg("source"),
+         "Deep copy data from source into this array, allocating as needed.")
+    .def("asnumpy",
+         &vip::ArrayHandleToNumPy,
+         "Return a read-only NumPy view of this array. Zero-copy for basic and\n"
+         "runtime-vec storage; other layouts are copied element-by-element.\n"
+         "The view keeps the source array alive for its entire lifetime.");
+
+  m.def("array_from_numpy",
+        &vip::NumPyToArrayHandle,
+        nb::arg("array"),
+        nb::arg("allow_copy") = false,
+        "Wrap a NumPy array as a Viskores UnknownArrayHandle. By default the\n"
+        "binding shares storage with the NumPy buffer; this requires the input\n"
+        "to be C-contiguous, aligned, and writable. Pass allow_copy=True to\n"
+        "let the binding make a contiguous, writable copy when the input\n"
+        "layout is not directly shareable. allow_copy is permission, not a\n"
+        "command: a directly-shareable input still takes the zero-copy path.");
+  m.def("asnumpy",
+        &vip::ArrayHandleToNumPy,
+        nb::arg("array"),
+        "Return a read-only NumPy view of a Viskores UnknownArrayHandle.\n"
+        "Equivalent to array.asnumpy().");
+}
+
+} // namespace viskores::python::bindings
--- a/python_bindings/testing/CMakeLists.txt
+++ b/python_bindings/testing/CMakeLists.txt
@@ -1,0 +1,23 @@
+##============================================================================
+##  The contents of this file are covered by the Viskores license. See
+##  LICENSE.txt for details.
+##
+##  By contributing to this file, all contributors agree to the Developer
+##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+##============================================================================
+
+# Viskores' module system processes this file in an isolated scope, so the
+# Python discovery from the parent python_bindings CMakeLists is not visible
+# here. Re-run it locally; find_package hits the CMake cache, so this is cheap
+# and does not depend on the parent's ordering.
+find_package(Python REQUIRED COMPONENTS Interpreter)
+
+add_test(
+  NAME PythonBindingsUnknownArrayHandleNumPy
+  COMMAND
+    "${Python_EXECUTABLE}"
+    "${CMAKE_CURRENT_SOURCE_DIR}/PythonBindingsUnknownArrayHandleNumPy.py"
+)
+set_tests_properties(PythonBindingsUnknownArrayHandleNumPy PROPERTIES
+  ENVIRONMENT "PYTHONPATH=${Viskores_BINARY_DIR}/python_bindings"
+)
--- a/python_bindings/viskores/__init__.py
+++ b/python_bindings/viskores/__init__.py
@@ -1,0 +1,9 @@
+##============================================================================
+##  The contents of this file are covered by the Viskores license. See
+##  LICENSE.txt for details.
+##
+##  By contributing to this file, all contributors agree to the Developer
+##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+##============================================================================
+
+__all__ = []
--- a/viskores/interop/python/ArrayHandleToNumPy.h
+++ b/viskores/interop/python/ArrayHandleToNumPy.h
@@ -1,0 +1,185 @@
+//============================================================================
+//  The contents of this file are covered by the Viskores license. See
+//  LICENSE.txt for details.
+//
+//  By contributing to this file, all contributors agree to the Developer
+//  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+//============================================================================
+
+// Convert a Viskores UnknownArrayHandle to a NumPy ndarray.
+// Requires nanobind (nanobind/ndarray.h) and Python development headers.
+
+#ifndef viskores_interop_python_ArrayHandleToNumPy_h
+#define viskores_interop_python_ArrayHandleToNumPy_h
+
+#include <viskores/interop/python/viskores_interop_python_export.h>
+
+#include <nanobind/nanobind.h>
+#include <nanobind/ndarray.h>
+
+#include <viskores/cont/ArrayHandleBasic.h>
+#include <viskores/cont/ArrayHandleRuntimeVec.h>
+#include <viskores/cont/Error.h>
+#include <viskores/cont/Token.h>
+#include <viskores/cont/UnknownArrayHandle.h>
+
+#include <memory>
+#include <type_traits>
+
+namespace nb = nanobind;
+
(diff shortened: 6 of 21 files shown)


python_bindings/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
+##============================================================================
+##  The contents of this file are covered by the Viskores license. See
+##  LICENSE.txt for details.
+##
+##  By contributing to this file, all contributors agree to the Developer
+##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+##============================================================================
+
+option(VISKORES_PYTHON_STABLE_ABI "Build Python bindings with the CPython stable ABI" OFF)
```
kmorel, line 11: Dumb question: what is the point of the stable ABI? Is that so a precompiled library can be distributed with a package manager?

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,647 @@
...
+  {
+    viskores::Vec<ComponentType, NumberOfComponents> value;
+    for (viskores::IdComponent component = 0; component < NumberOfComponents; ++component)
+    {
+      value[component] = values[(index * NumberOfComponents) + component];
+    }
+    portal.Set(index, value);
+  }
+  return result;
+}
```
kmorel: A better way to do this is to use [`ArrayHandleRuntimeVec`] to handle vecs of any component size.

Also, I disagree with the copying of arrays. The Viskores buffer object should hold a reference to a Python object so that they can share the data, and the array will keep around as long as either Python object or `ArrayHandle` exists.

[`ArrayHandleRuntimeVec`]: https://viskores.readthedocs.io/en/latest/memory-layout.html#runtime-vec-arrays

```suggestion
// Build an UnknownArrayHandle for a 2D NumPy array with fixed-width vector rows.
template <typename T>
viskores::cont::UnknownArrayHandle ArrayFromNumPyVectorArray(nb::object array)
{
  const viskores::Id numberOfValues = ShapeValue(array, 0);
  const viskores::IdComponent numberOfComponents = ShapeValue(array, 1);
  
  auto npDeleter = [](void* container){
    nb::handle* npContainer = reinterpret_cast<nb::handle*>(container);
    npContainer->reset();
    delete npContainer;
  };
  
  viskores::cont::ArrayHandleBasic<T> componentArray(
    PyBufferView(array).Data<T>(),
    new nb::object(array),
    numberOfValues * numberOfComponents,
    npDeleter);
    
  return viskores::cont::ArrayHandleRuntimeVec<T>(numberOfComponents, componentArray);
}
```

kmorel (changes requested):
I just did a quick look at a few things. I'm sure there are more to come, but the first thing that struck me is that we should avoid copying data between NumPy and Viskores. Both use reference counting on their arrays, so each should be able to hold a reference to the other.

ghweber:
@kmorel Yes, the main motivation for the stable ABI option is future binary distribution: an abi3 extension can be compatible across multiple CPython versions instead of being tied to one exact Python minor version. It is off by default here, so the normal build still uses the regular Python ABI.

A secondary reason for keeping the option this early is that it gives us a useful constraint while developing the bindings. If we accidentally rely on Python C API calls that are not available through the stable ABI, we can catch that before the code grows.

There is also a practical workflow motivation for me: I often use virtual environments for different projects, and sometimes those environments need different Python versions because of package compatibility issues. In that situation it is very helpful not to worry about rebuilding the Viskores Python binding for each Python minor version, and instead use one compiled binding where possible.

This PR is not trying to settle packaging policy yet. If the option feels like too much scope for the first array PR, we can split it out, but mechanically it is small and currently optional.

ghweber:
One concrete reason this matters to me: I have run into this with another compiled Python library, where it became a significant annoyance to rebuild the library in every custom virtual environment where I wanted to use it. Avoiding that kind of per-environment rebuild friction is exactly the type of use case where stable ABI support can help.

python_bindings/CMakeLists.txt
ghweber: The main motivation for the stable ABI option is future binary distribution: an abi3 extension can be compatible across multiple CPython versions instead of being tied to one exact Python minor version. It is off by default here, so the normal build still uses the regular Python ABI.

A secondary reason for keeping the option this early is that it gives us a useful constraint while developing the bindings. If we accidentally rely on Python C API calls that are not available through the stable ABI, we can catch that before the code grows.

There is also a practical workflow motivation for me: I often use virtual environments for different projects, and sometimes those environments need different Python versions because of package compatibility issues. In that situation it is very helpful not to worry about rebuilding the Viskores Python binding for each Python minor version, and instead use one compiled binding where possible.

This PR is not trying to settle packaging policy yet. If the option feels like too much scope for the first array PR, we can split it out, but mechanically it is small and currently optional.

ghweber:
@kmorel, now that the pin buffer PR is merged, this one is worth taking another look at.

python_bindings/examples/roundtrip_allow_copy.py
```diff
@@ -0,0 +1,39 @@
...
+
+import viskores
+
+
+# Demonstrate the allow_copy=True kwarg on viskores.cont.array_from_numpy.
+# By default the binding requires a directly-shareable input (C-contiguous,
+# aligned, writable) so it can wrap the NumPy buffer with zero copy. Inputs
+# that don't meet those requirements are rejected with a descriptive error.
+# Pass allow_copy=True to let the binding make a contiguous, writable copy
+# automatically when the input is not directly shareable.
```
kmorel, line 20: Eventually, the examples should be move to the user's guide with a chapter on Python bindings. However, we can delay that until we figure out the right way to insert Python code into the guide and have it compiled for testing purposes.

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,692 @@
...
+// `allowCopy` controls behavior when the input's layout cannot be shared
+// directly (non-contiguous, misaligned, or read-only):
+//   * allowCopy=false (default): raise an error. The caller keeps full control
+//     over allocations and avoids hidden copies.
+//   * allowCopy=true: produce a contiguous, writable copy and share storage
+//     with that copy. The original NumPy array is not connected to the
+//     resulting ArrayHandle, so subsequent mutations of the original do not
+//     propagate. A copy is only made when the input is not directly shareable;
+//     a shareable input still takes the zero-copy path.
+viskores::cont::UnknownArrayHandle ArrayFromNumPy(nb::handle object, bool allowCopy)
```
kmorel: I propose creating another module in `viskores\interop\python` that contains `NumPyToArrayHandle` and `ArrayHandleToNumPy`. The Python library could then link to those functions to create the binding, and other C++ code that manages Python objects could also use them.

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,692 @@
...
+template <typename ValueType>
+constexpr const char* NumPyDType()
+{
+  using ComponentType = std::remove_cv_t<ValueType>;
+#define VISKORES_PYTHON_NUMPY_DTYPE_CASE(Type, Name)                    \
+  if constexpr (std::is_same_v<ComponentType, Type>)                    \
+  {                                                                     \
+    return Name;                                                        \
+  }                                                                     \
+  else
```
kmorel: Let's remove the `else` from the end of this macro. It's correct the way it's used, but weird.


```suggestion
#define VISKORES_PYTHON_NUMPY_DTYPE_CASE(Type, Name)                    \
  if constexpr (std::is_same_v<ComponentType, Type>)                    \
  {                                                                     \
    return Name;                                                        \
  }
```

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,692 @@
...
+}
+
+// Own the Python-side lifetime state for a NumPy array shared with Viskores.
+//
+// Viskores stores only a raw data pointer in the ArrayHandleBasic buffer. This
+// owner is passed to Viskores as the buffer container so the NumPy array remains
+// alive while any Viskores buffer reference can still use that pointer.
+// Releasing the final Viskores buffer reference calls ReleasePythonBufferOwner,
+// which deletes this object, releases the Python buffer export, and drops the
+// Python reference to the NumPy array.
```
kmorel: This isn't quite right. The object gets deleted only after both `ReleasePythonBufferOwner` and all other references to the python object are deleted. The difference is important as the whole point is to ensure that Viskores won't delete buffers still in use by Python and vice versa.


```suggestion
// which drops the Python reference to the NumPy array, allowing it to be deleted
// once all other Python references are dropped.
```

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,692 @@
...
+}
+
+// Throw a runtime_error explaining why a NumPy array's layout cannot be shared.
+// Pre: reason != ShareableCheck::Shareable.
+[[noreturn]] void ThrowUnshareableNumPyArrayReason(ShareableCheck reason)
+{
+  switch (reason)
+  {
+    case ShareableCheck::NotCContiguous:
+      throw std::runtime_error(
```
kmorel: Unless there is a compelling reason not to, we should throw `viskores::cont::Error` subclasses.

Note that `viskores::cont::Error` inherits from `std::exception`, so it will still work with code checking for standard exceptions.

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,692 @@
...
+{
+  return nb::capsule(owner,
+                     [](void* pointer) noexcept { delete static_cast<OwnerType*>(pointer); });
+}
+
+// Return a zero-copy NumPy view of a basic ArrayHandle. The view owner pins
+// the host buffer (via Buffer::PinHost) so the pointer is locked for the
+// lifetime of the NumPy array.
+template <typename ValueType>
+nb::object BasicArrayToNumPy(const viskores::cont::ArrayHandleBasic<ValueType>& array)
```
kmorel: I don't see a reason why there should be both a `BasicArrayToNumPy` and a `RuntimeVecArrayToNumPy` function. You can pull a basic array out of an `UnknownArrayHandle` as a runtime vec array. So, just pull out a runtime vec all the time and handle the shape of one component as necessary.

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,692 @@
...
+    return result;
+  }
+
+  // For everything else, dispatch via CastAndCall. Basic storage takes the
+  // zero-copy view path; other storage layouts copy element-by-element.
+  // TypeListAll covers every scalar and fixed-width Vec type defined in
+  // viskores/TypeList.h; all have component types in the NumPy scalar table so
+  // the compile-time guard in NumPyDType<T> cannot fire for any type in this list.
+  unknown.CastAndCallForTypes<viskores::TypeListAll, VISKORES_DEFAULT_STORAGE_LIST>(
+    [&result](const auto& array) { result = ArrayHandleToNumPy(array); });
```
kmorel: I disagree with the cast and call here. If the attempt to pull the data as a runtime vec fails, you are unlikely to be able to find contiguous memory to shallow copy. Move to the deep copy method, which does not require a cast and call.

python_bindings/testing/src/test_helpers.cxx
```diff
@@ -0,0 +1,31 @@
...
+  // Produce a Viskores-allocated basic ArrayHandle with the same contents as
+  // the input. The result has no Python-side ownership of its buffer, so it
+  // exercises the `Buffer::PinHost` zero-copy path in `asnumpy`.
+  m.def("viskores_allocated_copy",
+        [](const viskores::cont::UnknownArrayHandle& src) {
+          viskores::cont::UnknownArrayHandle dest = src.NewInstanceBasic();
+          dest.DeepCopyFrom(src);
+          return dest;
+        },
+        nb::arg("array"));
```
kmorel: Rather than create an entire testing library to provide a deep copy function, why not just add a binding to the `DeepCopyFrom` method in `cont_bindings.cxx`?

python_bindings/testing/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+)
+# The test imports both `viskores` (public bindings) and `_viskores_test_helpers`,
+# so the public bindings target must exist on disk before the test runs.
+add_dependencies(viskores_python_test_helpers viskores_python_bindings)
+
+add_test(
+  NAME PythonBindingsUnknownArrayHandleNumPy
+  COMMAND
+    "${Python_EXECUTABLE}"
+    "${CMAKE_CURRENT_SOURCE_DIR}/test_unknown_array_numpy_roundtrip.py"
```
kmorel: To prevent confusion, make the test name and test source filename match.

python_bindings/CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
...
+)
+set(Viskores_PYTHON_PACKAGE_STAMP
+  "${Viskores_BINARY_DIR}/python_bindings/.python_package_sync.stamp")
+
+add_custom_command(
+  OUTPUT "${Viskores_PYTHON_PACKAGE_STAMP}"
+  COMMAND "${CMAKE_COMMAND}" -E make_directory "${Viskores_PYTHON_PACKAGE_ROOT}"
+  COMMAND "${CMAKE_COMMAND}" -E copy_directory
+          "${CMAKE_CURRENT_SOURCE_DIR}/viskores"
+          "${Viskores_PYTHON_PACKAGE_DIR}"
```
kmorel, line 96: This doesn't seem like it would add the proper installation for the install target.

kmorel (changes requested):

python_bindings/src/cont_bindings.cxx
ghweber: The `nb::object` approach is cleaner, but the current `PyBufferOwner` acquires a `Py_buffer` export via `PyObject_GetBuffer(PyBUF_WRITABLE)` which does more than keep the array alive: it signals to NumPy that there is an active consumer. While that export is held, NumPy prevents the user from resizing the array or changing its writability flag — operations that would silently invalidate the pointer Viskores is holding. A bare `nb::object` reference loses that guarantee.

@kmorel, is dropping the buffer protocol acceptable here, or do you prefer to keep it?

python_bindings/CMakeLists.txt
ghweber: Done. Added `install(TARGETS ...)` and `install(DIRECTORY ...)` rules that put the extension module and the Python source files under `lib/pythonX.Y/site-packages/viskores/` relative to `CMAKE_INSTALL_PREFIX`.

python_bindings/examples/roundtrip_allow_copy.py
ghweber: Agreed, deferred to a follow-up once we figure out how to embed and compile-test Python snippets in the user's guide.

python_bindings/src/cont_bindings.cxx
ghweber: Done. Extracted the conversion logic into two headers: `viskores/interop/python/ArrayHandleToNumPy.h` and `viskores/interop/python/NumPyToArrayHandle.h`. `cont_bindings.cxx` is now a thin binding layer (~70 lines) that includes them and registers the nanobind methods. Other C++ code that manages Python objects can include the interop headers directly.

python_bindings/src/cont_bindings.cxx
ghweber: Done, applied your suggested wording.

python_bindings/src/cont_bindings.cxx
ghweber: Done, switched to `viskores::cont::ErrorBadValue` throughout both conversion directions.

python_bindings/src/cont_bindings.cxx
ghweber: Done.

python_bindings/src/cont_bindings.cxx
ghweber: Done. The fallback is now `NewInstanceBasic() + DeepCopyFrom()` followed by a retry of the RuntimeVec view path. Removed `WritableBufferView`, `AllocateNumPyArray`, and the `CastAndCallForTypes` dispatch.

python_bindings/testing/src/test_helpers.cxx
ghweber: Done. `DeepCopyFrom` and `NewInstanceBasic` are now public methods on `UnknownArrayHandle` in `cont_bindings.cxx`. `test_helpers.cxx` and its CMake target are deleted; the test uses the public methods directly.

python_bindings/src/cont_bindings.cxx
ghweber: Done. Removed `BasicArrayToNumPy`, `ArrayHandleToNumPy<non-basic>`, and the scalar-exclusion guard. All arrays now go through `TryAnyRuntimeVecArrayToNumPy`; `RuntimeVecArrayToNumPy` emits a 1D view when `numberOfComponents == 1` so scalar Viskores arrays stay 1D in NumPy. `BasicArrayViewOwner` is also gone — `RuntimeVecArrayViewOwner` handles both shapes.

python_bindings/testing/CMakeLists.txt
ghweber: Done, renamed to `test_unknown_array_numpy_roundtrip`.

python_bindings/src/ArrayHandleToNumPy.h
```diff
@@ -0,0 +1,204 @@
...
+    viskores::cont::Token token;
+    this->ComponentsArray.GetBuffers()[0].PinHost(token);
+    this->Data = this->ComponentsArray.GetReadPointer(token);
+  }
+
+  const ComponentType* GetData() const { return this->Data; }
+
+private:
+  viskores::cont::ArrayHandleRuntimeVec<ComponentType> Array;
+  viskores::cont::ArrayHandleBasic<ComponentType> ComponentsArray;
```
kmorel: You only really need to hold a reference to either the `ArrayHandleRuntimeVec` or its `ArrayHandleBasic` components array. Both will hold a reference to the same `Buffer` and prevent it from going out of scope. (Of course, there is no harm in keeping both.)

viskores/interop/python/ArrayHandleToNumPy.h
```diff
@@ -0,0 +1,204 @@
...
+  viskores::cont::ArrayHandleBasic<ComponentType> ComponentsArray;
+  const ComponentType* Data = nullptr;
+};
+
+template <typename OwnerType>
+inline nb::capsule MakeArrayViewOwnerCapsule(OwnerType* owner)
+{
+  return nb::capsule(owner,
+                     [](void* pointer) noexcept { delete static_cast<OwnerType*>(pointer); });
+}
```
kmorel, line 113: I had to read the implementation several times and look up documentation for `std::unique_ptr` before I understood what it was doing. The code will be more clear by moving that defensive programming into `MakeArrayViewOwnerCapsule` and adding some helpful comments.


```suggestion
template <typename ComponentType>
inline nb::capsule MakeArrayViewOwnerCapsule(
  const viskores::cont::ArrayHandleRuntimeVec<ComponentType>& array)
{
  using OwnerType = RuntimeVecArrayViewOwner<ComponentType>;
  
  // Defensively create owner in a `std::unique_ptr` to defend against `nb:capsule`
  // throwing an exception.
  auto owner = std::make_unique<OwnerType>(array);
  nb::capsule ownerCapsule(owner.get(),
                           [](void* pointer) noexcept { delete static_cast<OwnerType*>(pointer); });

  // Owner pointer now safely manged by `ownerCapsule`. Release the pointer from the
  // `unique_ptr` without deleting it.
  owner.release();
  
  return ownerCapsule;
}
```

python_bindings/src/cont_bindings.cxx
kmorel: Is this done? I don't see a `viskores/interop/python` directory created.

kmorel (changes requested):
I got partway through reviewing this and realize that the new module `viskores/interop/python` that is supposed to be created is not there. Did these files not get added?

python_bindings/testing/CMakeLists.txt
kmorel: It would be better to change the source file to `PythonBindingsUnknownArrayHandleNumPy.py` rather than the test name. That way the test name will match the convention of the other tests in Viskores.

python_bindings/src/cont_bindings.cxx
ghweber: Done now — the files were misplaced in `python_bindings/src/` in the previous push. They are now at `viskores/interop/python/ArrayHandleToNumPy.h` and `viskores/interop/python/NumPyToArrayHandle.h`. Install rules in `python_bindings/CMakeLists.txt` put them in the install tree under `${Viskores_INSTALL_INCLUDE_DIR}/viskores/interop/python/`; the `SourceInInstall` test now passes.

python_bindings/testing/CMakeLists.txt
ghweber: Done. Renamed the source file to `PythonBindingsUnknownArrayHandleNumPy.py` and updated the CTest `NAME` and `set_tests_properties` call to match.

python_bindings/src/ArrayHandleToNumPy.h
ghweber: Simplified — `RuntimeVecArrayViewOwner` now holds only `ComponentsArray` (the `ArrayHandleBasic`). Added a comment explaining the invariant: `ArrayHandleBasic` and the `ArrayHandleRuntimeVec` it came from share the same `Buffer` via reference counting, so either one keeps it alive.

viskores/interop/python/ArrayHandleToNumPy.h
ghweber: Done. Changed the signature to `MakeArrayViewOwnerCapsule(std::unique_ptr<OwnerType> owner)` — taken by value so the `unique_ptr` guards the object inside the function until `nb::capsule` is successfully constructed. The call site saves `data = owner->GetData()` first, then passes `std::move(owner)` in.

ghweber:
The macOS and Windows CI builds were failing on `SourceInInstall`. That test scans every `*.h` under `viskores/` in the source tree and verifies each one appears in the install tree. Our new headers live at `viskores/interop/python/ArrayHandleToNumPy.h` and `viskores/interop/python/NumPyToArrayHandle.h` — inside the scanned tree — but the CI builds that don't build the Python bindings never processed `python_bindings/CMakeLists.txt`, so the install rules never ran.

The fix (commit `ab222a656`) follows the same pattern as `viskores/interop/internal/CMakeLists.txt`: a new `viskores/interop/python/CMakeLists.txt` calls `viskores_declare_headers()` to register the headers for installation unconditionally, and `viskores/interop/CMakeLists.txt` picks it up with `add_subdirectory(python)`. The now-redundant install block in `python_bindings/CMakeLists.txt` was removed.

@kmorel : Does this approach look right to you? The alternative would be to add `interop/python` to the `DIR_EXCEPTIONS` list in the CI configuration and keep the install rules inside the Python bindings gate — but that would mean the headers are only installed when Python bindings are built, which may not be what users expect. Installing them unconditionally means a vanilla Viskores install includes headers that `#include <nanobind/nanobind.h>`, even when nanobind isn't present; the headers can't be compiled without it, but the install itself is harmless. Happy to take this in a different direction if you prefer.

kmorel:
> @kmorel : Does this approach look right to you? The alternative would be to add `interop/python` to the `DIR_EXCEPTIONS` list in the CI configuration

I'm not a fan of any of these solutions. A better solution would be to make a module in the build system (using the `viskores.module` feature). This follows the approach of [`viskores/interop/anari`](https://github.com/Viskores/viskores/tree/main/viskores/interop/anari). For this, you need the following things:

* A `viskores/interop/python/viskores.module`
* Create `ArrayHandleToNumPy.cxx` and `NumPyToArrayHandle.cxx` source files for a library attached to this module. Change `UnknownArrayToNumPy` and `ArrayFromNumPy` from inline functions to exported library functions implemented in their respective .cxx files. (Also, the names of these functions should match the names of the source files. So change that.)
* Change `viskores/interop/python/CMakeLists.txt` to use the `viskores_library` CMake function to create the module library. This will automatically set up all install targetrs.
* Remove any `add_subdirectory` pointing to `viskores/interop/python`. The directory will be picked up with the `viskores.module` file.
* Update the python bindings module to depend on the python interop module (in `python_bindings/viskores.module`).

python_bindings/src/cont_bindings.cxx
kmorel: If I understand what you said correctly, when you create an ArrayHandle with the shared memory of a numpy array, the numpy array can no longer be resized.

I think this behavior is exactly correct (as long as numpy produces an appropriate error if an array resize is attempted). When going the other way, we pin the array from the ArrayHandle to make sure it cannot be resized and invalidate the numpy array. The same is true in the other way. Once you share a numpy array with Viskores, you are no longer allowed to reallocate the numpy array.

ghweber:
I've implemented the module restructuring Ken requested. Here's a summary of what changed and one design decision I'd like Ken's feedback on.

### What changed

**`viskores/interop/python/viskores.module`** (new)
Declares `viskores_interop_python` in the `PythonBindings` group with `NO_TESTING`. This makes the directory auto-excluded from `SourceInInstall` when `Viskores_ENABLE_PYTHON_BINDINGS` is OFF, which is what Ken pointed out was the right fix.

**`viskores/interop/python/CMakeLists.txt`**
Now uses `viskores_library` (static archive). The non-template entry points (`ArrayHandleToNumPy`, `NumPyToArrayHandle`) are compiled from `.cxx` files. nanobind and Python headers are provided at compile time via `target_include_directories`; the nanobind runtime is left to the consumer.

**`viskores/interop/python/ArrayHandleToNumPy.cxx`** / **`NumPyToArrayHandle.cxx`** (new)
Compiled implementations of the two entry points (bodies moved out of the headers).

**`viskores/interop/CMakeLists.txt`**
`add_subdirectory(python)` removed — the module system discovers the directory via `viskores.module`.

**`python_bindings/viskores.module`**
`viskores_interop_python` added as a `DEPENDS` entry.

**`python_bindings/CMakeLists.txt`**
`viskores_interop_python` added to `target_link_libraries` (required because `nanobind_add_module` bypasses the module system's automatic dependency linking).

---

### Design decision: static vs. shared library

I made `viskores_interop_python` a **static** archive (`viskores_library(... STATIC ...)`), not a shared library. The reason:

The build here has `VISKORES_PYTHON_STABLE_ABI=ON`, so `nanobind_add_module` links against `nanobind-static-abi3`. If `viskores_interop_python` were a separate shared library with its own copy of `nanobind-static` (a different variant), the two nanobind runtimes would conflict at startup — causing a segfault in practice.

Making it static means its object files are linked directly into `_viskores.abi3.so`, so both share exactly one nanobind runtime (whichever variant `nanobind_add_module` picks). The test suite passes with this approach.

An alternative would be to match the nanobind variant precisely (e.g., check `VISKORES_PYTHON_STABLE_ABI` in the interop CMakeLists and use `nanobind-static-abi3` or `nanobind-static` accordingly), but that's more coupling and the static approach is simpler and correct.

Ken — does a static `viskores_library` work for you, or would you prefer a different approach?

kmorel:
> Ken — does a static `viskores_library` work for you, or would you prefer a different approach?

I don't know enough about ABI and its issues to comment too much on them. I think it's OK to have that one library static. Make sure you have well documented why this particular library is declared static (likely in a comment in the CMakeLists.txt file).

@vicentebolea may know about some issues with static libraries that I do not.

python_bindings/src/ArrayHandleToNumPy.h
ghweber: Agreed — the current code stores only `ComponentsArray` (the `ArrayHandleBasic` extracted from the `RuntimeVec`), and the comment in `RuntimeVecArrayViewOwner` notes this explicitly. No change needed here.

python_bindings/src/cont_bindings.cxx
```diff
@@ -0,0 +1,78 @@
...
+void BindCont(nb::module_& m)
+{
+  nb::class_<viskores::cont::UnknownArrayHandle>(m, "UnknownArrayHandle")
+    .def(nb::init<>())
+    .def("__len__",
+         [](const viskores::cont::UnknownArrayHandle& self) { return self.GetNumberOfValues(); })
+    .def("__repr__", &UnknownArrayHandleRepr)
+    .def("GetNumberOfValues",
+         &viskores::cont::UnknownArrayHandle::GetNumberOfValues,
+         "Number of elements (tuples) in the array.")
```
kmorel: `__len__` and `GetNumberOfValues` appear to have the exact same functionality. Why is one implemented with a lambda and the other with a method reference?

python_bindings/viskores/__init__.py
```diff
@@ -0,0 +1,11 @@
+##============================================================================
+##  The contents of this file are covered by the Viskores license. See
+##  LICENSE.txt for details.
+##
+##  By contributing to this file, all contributors agree to the Developer
+##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+##============================================================================
+
+from . import cont
```
kmorel: This feels like it is implying that that with one

```python
import viskores
```

all viskores modules get loaded. As the Python bindings grow, this sounds like it could become problematic. It may make more sense to require loading each module individually, such as

```python
import viskores.cont
import viskores.filter.contour
import viskores.io
```

ghweber:
> > Ken — does a static `viskores_library` work for you, or would you prefer a different approach?
> 
> I don't know enough about ABI and its issues to comment too much on them. I think it's OK to have that one library static. Make sure you have well documented why this particular library is declared static (likely in a comment in the CMakeLists.txt file).
> 
> @vicentebolea may know about some issues with static libraries that I do not.

@vicentebolea could you let me know if there are any issues with static libraries?

kmorel (changes requested):
On my last look through the code, things are looking good to me. There was one point I addressed about not automatically loading submodules with `import viskores`. If you agree with that and change it, I think we are good.

python_bindings/src/cont_bindings.cxx
ghweber: Good catch — switched `__len__` to use the method reference directly, same as `GetNumberOfValues`.

python_bindings/viskores/__init__.py
ghweber: Agreed. Removed the `from . import cont` from `__init__.py` so `import viskores` no longer pulls in submodules. Users now need `import viskores.cont` explicitly.

ghweber:
Now that the Python tests pass locally, it would be good to run them in CI as well. I looked at the existing CI configuration and wanted to get input before making changes.

### Current state

Python is not present in any CI job:
- No dockerfile installs `python3-dev`, `nanobind`, or `numpy`
- `Viskores_ENABLE_PYTHON_BINDINGS` has no entry in `initial_config.cmake`'s settings parser

### What's needed

The changes are small regardless of approach:
1. **Dockerfile** — add `python3-dev`, `python3-pip`, `python3-numpy` to an image; install `nanobind` via pip (no distro package exists). ~4 lines.
2. **`initial_config.cmake`** — add a `python_bindings` settings token that sets `Viskores_ENABLE_PYTHON_BINDINGS=ON`. One `elseif` block.
3. **CI config** — add `python_bindings` to a `VISKORES_SETTINGS` string, or define a new job.

### Options

**Option A: Piggyback on an existing ubuntu2404 job** (e.g. `ubuntu2404_gcc14`)
Add `python3-dev`/nanobind/numpy to the ubuntu2404 dockerfile and add `python_bindings` to that job's `VISKORES_SETTINGS`. Minimal CI overhead — no new job, no new image. Downside: adds Python build+test time to a job that currently doesn't have it.

**Option B: Add a dedicated `ubuntu2404_python` job**
New job extending `.ubuntu2404` with `VISKORES_SETTINGS: "ccache+openmp+python_bindings"`. Keeps Python testing isolated and makes failures easy to attribute. Requires the same dockerfile change since it uses the same base image.

**Option C: New dockerfile + job**
A separate minimal image (no ANARI, no vtk-types, etc.) with Python dev tools. Most isolated but most work, and requires a new image to be built and pushed.

My preference would be **Option A or B** — both require only the dockerfile change plus a few lines of CMake/CI config, and the ubuntu2404 image is already the most up-to-date one. A separate job (Option B) seems cleaner since it won't inflate the runtime of an existing job.

@kmorel @vicentebolea — does this approach sound right? Is there a preferred image/job to target, or a CI convention I should follow for optional-dependency jobs like this?

kmorel:
> Now that the Python tests pass locally, it would be good to run them in CI as well. I looked at the existing CI configuration and wanted to get input before making changes.

I propose we check this in as an experimental feature without the CI testing enabled yet. Then, before any further Python binding PRs are submitted, we have a PR that adds the Python testing to the CI. (I'm a big fan of small PRs.)

> My preference would be **Option A or B**

I think we should do option A. I also think we should have at least two CI jobs test the python bindings. At least one should be CPU-only and one should use the GPU.

> does this approach sound right?

That's the basic idea. The part that is missing is that you have to build an image for the Dockerfile changes, add that image to the Docker registry, and update `.gitlab-ci.yml` to point to the new image. See #348 for an example of doing this.

I did #348 with Codex. It did a good job modifying the Dockerfile, building a new image, and registering it. @vicentebolea may have helped me rebuild the image since my Mac doesn't have an x86 processor. I can't remember if that part worked. Still, those issues are easier to resolve if you have a small PR that just changes the tests.

kmorel:
Do your examples still work? It looks like they are using `viskores.cont` without importing it, which is required now.

Otherwise the code looks good. Fix the examples, [squash the commits](https://github.com/Viskores/viskores/blob/main/CONTRIBUTING.md#squash-commits), and I think this is ready to be merged.

ghweber:
Fixed the examples and squashed all commits into one. CI is running on the force-pushed branch. Could you approve the review when you get a chance?

kmorel (approved):
Looks good. Once the CI passes, this is ready to merge.

PR #309, opened 2026-05-05
ci: use aarch64 runner for spack sync job

vicentebolea:


vicentebolea:

vicentebolea:
@kmorel I have temporary made the sync script to use the arm64 runner while the x86_64 runners are unpaused. This should allow triggering the Frank CI pipeline.

PR #310, opened 2026-05-06
Cell locator id only changelog

dpugmire:
Forgot to add the change log for #260.


vicentebolea (approved):

kmorel:
Does this need to be backported to the release branch?

dpugmire:
yes, it should be tied to #260. Did 260 make it in?

kmorel:
Never mind. I didn't realize #260 was only merged yesterday, so it's not in the release branch.

If you want to get this into the release, we can still do that. (The release is waiting for another fix as well.) In that case, we would backport both.

PR #316, opened 2026-05-15
Isolate contour tree filter headers from device code

kmorel:
Some of the contour tree filter headers included internal classes that exposed device (execution environment) code. Filters should not do that so software using the filters can avoid compiling with the GPU compiler if they otherwise do not need it. The contour tree filter code is updated to avoid such extraneous includes.


kmorel:
@ghweber Please review. Although there a lot of lines, most of them are moving code from a .h to a .cxx.

viskores/filter/scalar_topology/worklet/contourtree_augmented/Types.h
```diff
@@ -195,8 +198,7 @@ VISKORES_CONT inline void AssertArrayHandleNoFlagsSet(
 VISKORES_CONT
 inline void IdArraySetValue(viskores::Id index, viskores::Id value, IdArrayType& arr)
```
ghweber: Given that this now just calls/replicates `viskores::cont::ArraySetValue` I am wondering if we should just remove this function and call `viskores::cont::ArraySetValue` directly.

viskores/filter/scalar_topology/worklet/contourtree_augmented/Types.h
```diff
@@ -212,27 +214,8 @@ void ResizeVector(viskores::cont::ArrayHandle<ValueType>& thearray,
...
-    viskores::cont::Algorithm::CopySubRange(
-      viskores::cont::ArrayHandleConstant<ValueType>(fillValue, newSize - oldSize), // copy
-      0,                 // start copying from first index
-      newSize - oldSize, // num values to copy
-      thearray,          // target array to copy to
-      oldSize            // start copy to after oldSize
-    );
-  }
+  // Resize the array but keep the original values. Fill expanded with provided value
+  thearray.AllocateAndFill(newSize, fillValue, viskores::CopyFlag::On);
```
ghweber: Same as previous comment. Wondering if we should just eliminate this function and use `AllocateAndFill` directly.

viskores/filter/scalar_topology/ContourTreeUniformAugmented.h
```diff
@@ -117,6 +128,10 @@ class VISKORES_FILTER_SCALAR_TOPOLOGY_EXPORT ContourTreeAugmented : public visko
   explicit ContourTreeAugmented(bool useMarchingCubes = false,
                                 unsigned int computeRegularStructure = 1);
 
+  // Required for incomplete type MultiBlockContourTreeHelper
+  ContourTreeAugmented(ContourTreeAugmented&& src);
+  ~ContourTreeAugmented();
+
```
ghweber, line 134: Comment by AI: Adding an out-of-line destructor and move constructor fixes the incomplete `unique_ptr` type, but it also removes implicit move assignment. Code like
```
    a = std::move(b)
```
 likely worked before and becomes ill-formed. Add a defaulted out-of-line move assignment operator as well, preferably noexcept.

Personally, I don't think this is an issue we need to worry about as I don't forsee needing to use this, but on the other hand it may come back to bite us. However, I also think a longer term solution is to clean up the contour tree filter to avoid storing results in the filter and instead passing them as output data. But that change goes beyond the scope of this PR.

viskores/filter/scalar_topology/ContourTreeUniformDistributed.h
```diff
@@ -224,25 +225,12 @@ class VISKORES_FILTER_SCALAR_TOPOLOGY_EXPORT ContourTreeUniformDistributed
...
-  std::vector<viskores::worklet::contourtree_distributed::InteriorForest> LocalInteriorForests;
-
-  /// The hierarchical trees computed by the filter (array with one entry per block)
-  // TODO/FIXME: We need to find a way to store the final hieararchical trees somewhere.
-  // Currently we cannot do this here as it is a template on FieldType
-  //
-  //std::vector<viskores::worklet::contourtree_distributed::HierarchicalContourTree> HierarchicalContourTrees;
-  /// Number of iterations used to compute the contour tree
-  viskores::Id NumIterations;
+  struct InternalStruct;
```
ghweber, line 232: Analyzing this jointly with an LLM found a potential issue here: "the new `std::shared_ptr<InternalStruct>` makes copied filter instances share mutable execution scratch (`LocalMeshes`, `LocalContourTrees`, etc.  in `.cxx:104-120`). Before this PR, copies got independent containers. A copied filter now can race or cross-contaminate state if two copies execute independently. I’d either deep-copy/reset Internals in copy construction/assignment, or make the filter explicitly move-only if copy is not supported.`"

A quick fix may be using a `unique_ptr` analogously to `ContourTreeUniformAugmented`.

However, this scratch space is only needed during the computation. If it is permissible to change the signature of `PreExecute` and `DoPostExecute` then a better option would likely be to pass an `ExecutionState` (or similar) object. (I thought those were Viskores functions with fixed signature, but that no longer seems to be the case, so changing signatures may be ok).

This could look as follows:
* Rename `InternalStruct` to `ExecutionState` or similar.
* Have functions as follows:
```
class ContourTreeUniformDistributed : public viskores::filter::Filter
  {
  private:
    struct ExecutionState;

[...]

    VISKORES_CONT void PreExecute(
      const viskores::cont::PartitionedDataSet& input,
      ExecutionState& state);

    template <typename T, typename StorageType>
    VISKORES_CONT void ComputeLocalTree(
      viskores::Id blockIndex,
      const viskores::cont::DataSet& input,
      const viskores::cont::ArrayHandle<T, StorageType>& fieldArray,
      ExecutionState& state);

    template <typename T, typename StorageType, typename MeshType, typename MeshBoundaryExecType>
    VISKORES_CONT void ComputeLocalTreeImpl(
      viskores::Id blockIndex,
      const viskores::cont::DataSet& input,
      const viskores::cont::ArrayHandle<T, StorageType>& field,
      MeshType& mesh,
      MeshBoundaryExecType& meshBoundaryExecObject,
      ExecutionState& state);

    template <typename FieldType>
    VISKORES_CONT void DoPostExecute(
      const viskores::cont::PartitionedDataSet& input,
      viskores::cont::PartitionedDataSet& output,
      ExecutionState& state);

[ ...]

  };
```

And then something like:
```
  viskores::cont::PartitionedDataSet
  ContourTreeUniformDistributed::DoExecutePartitions(
    const viskores::cont::PartitionedDataSet& input)
  {
    ExecutionState state;

    this->PreExecute(input, state);

    for (viskores::Id blockNo = 0; blockNo < input.GetNumberOfPartitions(); ++blockNo)
    {
      const auto& dataset = input.GetPartition(blockNo);
      const auto& field =
        dataset.GetField(this->GetActiveFieldName(), this->GetActiveFieldAssociation());

      this->CastAndCallScalarField(
        field,
        [&](const auto& concrete)
        {
          this->ComputeLocalTree(blockNo, dataset, concrete, state);
        });
    }

    viskores::cont::PartitionedDataSet result;

    auto field =
      input.GetPartition(0).GetField(this->GetActiveFieldName(), this->GetActiveFieldAssociation());

    this->CastAndCallScalarField(
      field,
      [&](const auto& concrete)
      {
        using FieldType = typename std::decay_t<decltype(concrete)>::ValueType;
        this->DoPostExecute<FieldType>(input, result, state);
      });

    return result;
  }
```

And use the passed parameter `state` instead of `this->internals` in the functions that need to access the state.

kmorel: I partially agree with this. `unique_ptr` would be more appropriate here. I am being lazy so I did not have to create custom copy constructors/operators.

However, I argue that using `shared_ptr` here is no worse than what was there before. The state wrapped in the `shared_ptr` contains several `ArrayHandle` objects, and these behave just like a shared pointer. So, the problems you point out were already there before this change and would still exist even if I changed this to `unique_ptr` such as with the [`LocalBlockIndices`](https://github.com/kmorel/viskores/blob/36cb203bc92dcb2d1d594470d23571e66479efec/viskores/filter/scalar_topology/ContourTreeUniformDistributed.h#L230) declared right above it.

I agree that the best fix would be to remove intermediate state from the class members and modify the call specification of the execute methods like you describe. But this is **_way_** outside the scope of this PR (and, frankly really low on my priority list).

So, if it is OK, I would prefer to leave this as is, and you can fix the larger issue at a later date.

viskores/filter/scalar_topology/worklet/contourtree_augmented/Types.h
kmorel: Done.

viskores/filter/scalar_topology/worklet/contourtree_augmented/Types.h
kmorel: Done.

viskores/filter/scalar_topology/ContourTreeUniformDistributed.h
ghweber: I think it should be straightforward to remove intermediate state from the class members. I can either try to do a commit for this on this PR (if that is allowed during the regular review process) or try to do a new one once this one is merged.

kmorel: I insist that it happen on a different PR. Each PR should have the smallest reasonable scope, and this change is nearly orthogonal to the intention of this PR.

ghweber: Ok. Makes sense.

ghweber (approved):

PR #319, opened 2026-05-18
Add id-only BoundsMap locator for flow

dpugmire:
Adds a CellLocatorUniformBins-based block-bounds locator to BoundsMap for flow filters.

Updated the tests as well.
#258 was getting too large, so I'm splitting this into multiple MRs.

### DIFF
--- a/docs/changelog/flow-boundsmap-id-locator.md
+++ b/docs/changelog/flow-boundsmap-id-locator.md
@@ -1,0 +1,6 @@
+## Use id-only block lookup for flow bounds
+
+Flow filters now build an internal block-bounds locator that can identify all
+blocks containing a particle position using cell-id-only locator queries. This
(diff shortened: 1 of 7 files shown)


kmorel (changes requested):
Thanks for breaking up this work. That helps a lot.

Overall, this PR looks good. The one change I request is to move the new `BoundsMap` testing from `UnitTestStreamlineFilter.cxx` to its own test. That will keep the testing of the filter interface there and the internals of advection elsewhere. Plus, if you create a new test you can compile it for devices so you don't have to jump through hoops with the serial device adapter.

kmorel (approved):
Looks good once CI passes.

dpugmire:
@vicentebolea Are some of the test machines stuck? A few of these tests have been pending for a long time.

PR #325, opened 2026-05-20
Fix UnitTestAtomic coverage

dpugmire:
Fixes `UnitTestAtomic` so each atomic operation test actually invokes the matching functor and validates the expected result.

Also adds deterministic coverage for atomic return values, checking that `AtomicAdd`, `AtomicAnd`, `AtomicOr`, `AtomicXor`, and `AtomicNot` return the old stored value as documented.

## Details

- Correct `TestOr`, `TestXor`, `TestNot`, and `TestCompareExchange` to call the intended functors.
- Fix OR/XOR bit masks by casting to `T` before shifting, preserving `UInt64` high-bit coverage.
- Correct `TestNot` expected values for indices beyond the overlap count.
- Add `AtomicReturnValues` coverage for old-value return semantics.

### DIFF
--- a/viskores/testing/UnitTestAtomic.cxx
+++ b/viskores/testing/UnitTestAtomic.cxx
@@ -31,6 +31,8 @@
 
 #include <viskores/cont/testing/Testing.h>
 
+#include <vector>
+
 namespace
 {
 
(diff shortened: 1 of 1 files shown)


viskores/testing/UnitTestAtomic.cxx
```diff
@@ -123,6 +123,55 @@ struct AtomicTests
...
+      this->CheckEqual(viskores::AtomicNot(value), T(0x6), "Bad AtomicNot return value");
+    }
+  };
+
+  VISKORES_CONT void TestReturnValues()
+  {
+    std::cout << "AtomicReturnValues" << std::endl;
+    viskores::cont::ArrayHandleBasic<T> array;
+    viskores::cont::ArrayCopy(viskores::cont::make_ArrayHandleConstant<T>(T(0x6), ARRAY_SIZE),
+                              array);
```
kmorel:
```suggestion
    array.AllocateAndFill(ARRAY_SIZE, T(0x6));
```

kmorel (changes requested):
Just some minor changes. First, let's favor using `AllocateAndFill` over the more verbose array copy of constant array to initialize an array. Maybe with enough examples LLMs will start getting that right.

Also, please add a changelog sumarizing the improvements in testing coverage. It doesn't have to be too specific, so it can cover the several PRs you are submitting.

One more optional thing that is harder but would be nice. I notice that the testing of the return value does not actually test the atomic nature because only one thread operates on each value. It would be a better test if all threads did an atomic operation on the same value in memory. This is harder because the opeartions happen in an indeterminant order, but could be checked by stashing the return values in an output field array and then checking later. For example, you could test `AtomicAdd` by having all N threads add 1 to the same memory location. The result of the added value should be N and the output array should have the values [0..N-1] in random order. Sort the array and check. If you did the same thing with `AtomicXor`, you would get half the values in the output array as 0 and half as 1.

That said, I don't want to have perfection get in the way of improvement, so I am fine with merging this without those changes.

kmorel:
BTW, my notes about checking add and xor were examples. You should be able to check all atomics.

* `AtomicNot`: Basically the same as `AtomicXor`. Half of the values should be 0 and the other half... whatever the not of 0 is.
* `AtomicOr`: Start with 0 and or the value 1 << WorkId. If you count the 1 bits in each array, you should get the values from 0 to N-1.
* `AtomicAnd`: Similar to `AtomicOr`. Start with the value `~0` (all 1 bits). And the value ~(1 << WorkId). Then count the 0 bits in the resulting array.

dpugmire:
Yes, I planned to add more if these all passed on the cuda tests, etc.

kmorel (approved):
Looks good!

PR #326, opened 2026-05-20
Workletize flow seed block routing

dpugmire:
`AdvectAlgorithm::SetSeeds` previously performed a host-side bounds scan for every seed and immediately stored the results in nested `std::vector`s. This moves candidate discovery, validation, ordering, ownership filtering, and compaction to Viskores worklets and algorithms while preserving the current host-side queue interface.

Main changes include:
- route flow seeds to candidate blocks with `BoundsMap` locator worklets
- validate locator candidates against the original `Float64` block bounds
- deterministically order overlapping candidates and filter seeds by primary-rank ownership
- keep compacted particles and candidate groups in `ArrayHandle` representations until the existing host queue handoff
- add coverage for overlapping bounds, empty seeds, zero blocks, compaction, sorting, and coordinate precision

Note: this is a change that will be part of a multi-PR effort to workletize the streamline code.

### DIFF
--- a/viskores/filter/flow/internal/ParticleBlockIds.h
+++ b/viskores/filter/flow/internal/ParticleBlockIds.h
@@ -1,0 +1,341 @@
+//============================================================================
+//  The contents of this file are covered by the Viskores license. See
+//  LICENSE.txt for details.
+//
+//  By contributing to this file, all contributors agree to the Developer
+//  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+//============================================================================
+
+#ifndef viskores_filter_flow_internal_ParticleBlockIds_h
+#define viskores_filter_flow_internal_ParticleBlockIds_h
+
+#include <viskores/cont/Algorithm.h>
+#include <viskores/cont/ArrayHandle.h>
+#include <viskores/cont/ArrayHandleGroupVecVariable.h>
+#include <viskores/cont/ArrayHandleIndex.h>
+#include <viskores/cont/ArrayHandlePermutation.h>
+#include <viskores/cont/ConvertNumComponentsToOffsets.h>
+#include <viskores/cont/Invoker.h>
+#include <viskores/filter/flow/internal/BoundsMap.h>
+#include <viskores/worklet/WorkletMapField.h>
+
+#include <utility>
+#include <vector>
+
+namespace viskores
+{
+namespace filter
+{
+namespace flow
+{
+namespace internal
+{
+
+using BlockIdArrayHandle = viskores::cont::ArrayHandle<viskores::Id>;
+
+// Variable-length candidate block IDs for seeds. Each value is the group of block IDs
+// associated with one seed, backed by flat block ID storage and an offsets array.
+using CandidateBlockIdArrayHandle =
+  viskores::cont::ArrayHandleGroupVecVariable<BlockIdArrayHandle, BlockIdArrayHandle>;
+
+// A view of the candidate groups selected for this rank. The permutation keeps the
+// selected groups in seed order without copying their variable-length block ID storage.
+using SelectedCandidateBlockIdArrayHandle =
+  viskores::cont::ArrayHandlePermutation<BlockIdArrayHandle, CandidateBlockIdArrayHandle>;
+
+// Routing data for the seeds retained by this rank. The arrays are index-aligned,
+// and each particle's nonempty candidate group starts with its primary block ID.
+template <typename ParticleType>
+struct SeedBlockRoutingResult
+{
+  viskores::cont::ArrayHandle<ParticleType> Particles;
+  SelectedCandidateBlockIdArrayHandle CandidateBlockIds;
+};
+
+namespace detail
+{
+
+class CountParticleBlockIdsWorklet : public viskores::worklet::WorkletMapField
+{
+public:
+  using ControlSignature = void(FieldIn particles, ExecObject locator, FieldOut count);
+  using ExecutionSignature = void(_1, _2, _3);
+
+  template <typename ParticleType, typename LocatorType>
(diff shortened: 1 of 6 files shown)


dpugmire:
@kmorel This is ready for review.

viskores/filter/flow/internal/ParticleBlockIds.h
```diff
@@ -0,0 +1,336 @@
...
+    viskores::cont::make_ArrayHandleGroupVecVariable(candidateIds, candidateOffsets);
+
+  if (totalCandidateIds > 0)
+    invoker(detail::FindParticleBlockIdsWorklet{},
+            particles,
+            boundsMap.GetLocator(),
+            groupedCandidateIds);
+
+  if (totalCandidateIds == 0)
+    return groupedCandidateIds;
```
kmorel: I asked Codex to review this PR, and it flagged this section of code with this issue (which I never would have found myself).

> Seeds in zero-dimensional block bounds are incorrectly discarded
>
> Routing now relies exclusively on CellLocatorUniformBins::CountAllCells for broad-phase candidate discovery. BoundsMap represents a block whose bounds collapse to a single point as a vertex cell, but the locator returns zero candidates at that exact point. Consequently, FindParticleBlockIds returns an empty group and RouteSeedsToBlocks drops the seed.
>
> The previous BoundsMap::FindBlocks implementation directly called Bounds::Contains, so it retained the seed. I reproduced the regression by querying the existing vertex bounds (24, 0, 0) in TestBoundsMapLocatorHandlesDegenerateBounds; CountAllCells returned 0 instead of block ID 6.
>
> Candidate discovery needs to support vertex bounds, or such blocks must be checked through a fallback path. A regression test should exercise FindParticleBlockIds or RouteSeedsToBlocks with a point-bounds block.

I don't follow all the details of this, but I do notice that the function `TestBoundsMapLocatorHandlesDegenerateBounds` in `UnitTestBoundsMap.cxx` declares bounds for a vertex region but does not actually test it. I would suggest either enabling that test and fixing any issues or be clear about not supporting it.

viskores/filter/flow/testing/UnitTestBoundsMap.cxx
kmorel, line 427: Here is the aforementioned test for degenerate bounds. A 0D bounds is declared in the bounds vector, but that bounds is never tested. Perhaps it should be.

```suggestion
auto cellIds = FindBoundsMapCellIds(boundsMap,
{ viskores::Vec3f(1.0f, 1.0f, 0.0f),
viskores::Vec3f(5.0f, 0.0f, 1.0f),
viskores::Vec3f(8.0f, 1.0f, 1.0f),
viskores::Vec3f(13.0f, 0.0f, 0.0f),
viskores::Vec3f(16.0f, 1.0f, 0.0f),
viskores::Vec3f(20.0f, 0.0f, 1.0f),
viskores::Vec3f(24.0f, 0.0f, 1.0f) });

ValidateCellIds(cellIds[0], { 0 }, "BoundsMap locator failed for XY bounds");
ValidateCellIds(cellIds[1], { 1 }, "BoundsMap locator failed for XZ bounds");
ValidateCellIds(cellIds[2], { 2 }, "BoundsMap locator failed for YZ bounds");
ValidateCellIds(cellIds[3], { 3 }, "BoundsMap locator failed for X line bounds");
ValidateCellIds(cellIds[4], { 4 }, "BoundsMap locator failed for Y line bounds");
ValidateCellIds(cellIds[5], { 5 }, "BoundsMap locator failed for Z line bounds");
ValidateCellIds(cellIds[6], { 6 }, "BoundsMap locator failed for Z line bounds");
```

viskores/filter/flow/internal/ParticleBlockIds.h
```diff
@@ -0,0 +1,336 @@
...
+  }
+
+private:
+  viskores::Id Rank;
+};
+
+} // namespace detail
+
+template <typename ParticleType, typename StorageType>
+VISKORES_CONT auto FindParticleBlockIds(
```
kmorel: I know it's common practice to use `auto` a lot and even encouraged by folks like Meyers, but I find it frustrating keeping track of variable types that I have to trace through multiple function calls. If a type is supposed to be `CandidateBlockIdArrayHandle`, just declare it as such. That makes it easier for code reviews and allows the compiler to help you identify when your type unexpectantly changes.

viskores/filter/flow/internal/ParticleBlockIds.h
```diff
@@ -0,0 +1,336 @@
...
+  const viskores::filter::flow::internal::BoundsMap& boundsMap)
+{
+  // A particle can be inside multiple overlapping blocks. Count first so the
+  // variable-length block id groups can be allocated exactly, then fill them.
+  viskores::cont::Invoker invoker;
+  viskores::cont::ArrayHandle<viskores::IdComponent> blockCounts;
+  if (boundsMap.GetTotalNumBlocks() > 0)
+    invoker(detail::CountParticleBlockIdsWorklet{}, particles, boundsMap.GetLocator(), blockCounts);
+  else
+    blockCounts.AllocateAndFill(particles.GetNumberOfValues(), viskores::IdComponent{ 0 });
```
kmorel: If you know none of the particles are in a block, couldn't you just initialize the result directly and return that?

```suggestion
  else
  {
    BlockIdArrayHandle offsets;
    offsets.AllocateAndFill(particles.GetNumberOfValues() + 1, 0);
    return viskores::cont::make_ArrayHandleGroupVecVariable(BlockIdArrayHandle{}, offsets);
  }
```

kmorel (changes requested):
Codex identified something that might need to be fixed. My other comments are cosmetic.

dpugmire:
@kmorel thanks for the review. Does this look ok now?

kmorel (approved):
Looks good.

PR #332, opened 2026-05-27
Add quad geometry to ANARI device.

dpugmire:
This adds a new quad geometry implementation modeled after the existing triangle geometry path. The device now advertises `khr_geometry_quad`, accepts ANARI `quad` geometry with `vertex.position` and optional `primitive.index`, builds a Viskores quad cell set, and renders it through the existing Viskores quad raytracing extractor/intersector.

The change also adds an image-based ANARI quad geometry test. The test renders an indexed cube made from quad faces, with per-face coloring through `primitive.attribute0` and an `image1D` sampler, then compares against a new baseline image in `viskores-data`.

### DIFF
--- a/viskores/interop/anari/testing/CMakeLists.txt
+++ b/viskores/interop/anari/testing/CMakeLists.txt
@@ -22,5 +22,6 @@ SOURCES
   UnitTestANARIMapperPoints.cxx
   UnitTestANARIMapperTriangles.cxx
   UnitTestANARIMapperVolume.cxx
+  UnitTestANARIGeometryQuad.cxx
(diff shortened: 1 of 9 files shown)


kmorel:
It looks like the [new test results](https://open.cdash.org/tests/2579670890) suggest that the helide device gives significantly different results than the viskores device. You should probably look into that.

BTW, I am working on testing ANARI on both helide and viskores devices to ensure some consistency. It is just not there yet.

kmorel (approved):
Looks good once all tests pass.

Be sure to merge https://github.com/Viskores/viskores-data/pull/8 once this is merged.

PR #334, opened 2026-05-27
Add ANARI device load function

kmorel:
The ANARI interop library now contains a function named `ANARILoadDevice` that finds an ANARI library of a given name and loads its device. It also has a special feature that detects when loading the "viskores" library and creates the device from direct linking.


kmorel:
@jeffamstutz Please review. This incorporates a loading mechanism discussed in #267.

jeffamstutz (approved):
LGTM, nice addition.

vicentebolea (approved):
Looks good! Factory free functions seems to be a more succinct approach compared to the full class added in the last pR

PR #338, opened 2026-05-29
Fix clip edge interpolation

### LINKED ISSUE #330: ClipWithImplicitFunction creates cells that extend outside of input after wedge vertex reordering PR #287
It seems that  #287 broke the `ClipWithImplicitFunction` filter. I tested with a `CellSetExplicit` with a single unit-cube cell and clipped away more than half of the cube. The resulting single wedge extends beyond the input cube.

Copilot created this small program to check this:

[clip.cpp](https://github.com/user-attachments/files/28141298/clip.cpp)

When compiled against 32a07e04952d5a7f038b69ebbc68668793307ac6, the output is:
```
AABB of clipped result:
  Min: (0, -0.25, 0)
  Max: (1.25, 1, 1)
```
Whereas when built against b1641b0d30902002e1fe4f1284f68128b45d73a5, the out put is as expected:
```
AABB of clipped result:
  Min: (0, 0.25, 0)
  Max: (0.75, 1, 1)
```

spyridon97:
This PR resolves https://github.com/Viskores/viskores/issues/330.

### DIFF
--- a/viskores/filter/contour/testing/UnitTestClipWithFieldFilter.cxx
+++ b/viskores/filter/contour/testing/UnitTestClipWithFieldFilter.cxx
@@ -78,7 +78,7 @@ void TestClipExplicit()
   viskores::cont::ArrayHandle<viskores::Float32> resultArrayHandle;
   temp.AsArrayHandle(resultArrayHandle);
 
-  viskores::Float32 expected[6] = { 1, 2, 1, 1.5, 3.5, 1.5 };
+  viskores::Float32 expected[6] = { 1, 2, 1, 0.5, 0.5, 0.5 };
   for (int i = 0; i < 6; ++i)
   {
     VISKORES_TEST_ASSERT(test_equal(resultArrayHandle.ReadPortal().Get(i), expected[i]),
--- a/viskores/filter/contour/worklet/Clip.h
+++ b/viskores/filter/contour/worklet/Clip.h
@@ -405,14 +405,14 @@ public:
                 ei.Vertex1 = points[edge[0]];
                 ei.Vertex2 = points[edge[1]];
                 auto point1ToPoint2 =
-                  static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - scalars.Get(ei.Vertex2));
+                  static_cast<viskores::Float64>(scalars.Get(ei.Vertex2) - scalars.Get(ei.Vertex1));
                 if (point1ToPoint2 < 0)
                 {
                   viskores::Swap(ei.Vertex1, ei.Vertex2);
                   point1ToPoint2 = -point1ToPoint2;
                 }
                 auto point1ToIso =
-                  0.0 - (static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - this->IsoValue));
+                  0.0 - static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - this->IsoValue);
                 ei.Weight = point1ToPoint2 != 0 ? point1ToIso / point1ToPoint2 : 0;
                 // swap because edges are expected to be smallest,largest, t
                 if (ei.Vertex1 > ei.Vertex2)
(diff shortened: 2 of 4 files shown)


viskores/filter/contour/testing/UnitTestClipWithFieldFilter.cxx
```diff
@@ -78,7 +78,7 @@ void TestClipExplicit()
   viskores::cont::ArrayHandle<viskores::Float32> resultArrayHandle;
   temp.AsArrayHandle(resultArrayHandle);
 
-  viskores::Float32 expected[6] = { 1, 2, 1, 1.5, 3.5, 1.5 };
+  viskores::Float32 expected[6] = { 1, 2, 1, 0.5, 0.5, 0.5 };
```
kmorel, line 81: Hmm. It probably should have been a clue that the "expected" values of the field were outside the range of the input field values. (And the created points did not equal the isovalue.)

viskores/filter/contour/worklet/Clip.h
```diff
@@ -405,14 +405,14 @@ class Clip
...
+                  static_cast<viskores::Float64>(scalars.Get(ei.Vertex2) - scalars.Get(ei.Vertex1));
                 if (point1ToPoint2 < 0)
                 {
                   viskores::Swap(ei.Vertex1, ei.Vertex2);
                   point1ToPoint2 = -point1ToPoint2;
                 }
                 auto point1ToIso =
-                  0.0 - (static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - this->IsoValue));
+                  0.0 - static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - this->IsoValue);
                 ei.Weight = point1ToPoint2 != 0 ? point1ToIso / point1ToPoint2 : 0;
```
kmorel, line 416: Can you explain this fix? The comment in the changelog suggests that the problem was when the scalar values were the same on both sides of the edge. But in that case, `point1ToPoint2 == 0` (both before and after the fix). That would mean that `ei.Weight` would get `0` both before and after this change.

I'm also surprised by this fix. I would expect the correct `point1ToIso` to be `this->IsoValue - scalars.Get(ei.Vertex1)`, which is what it was originally doing (written in a really bizzare order).

viskores/filter/contour/testing/UnitTestClipWithFieldFilter.cxx
spyridon97: It should have been a clue, but i did not do the math back then because i thought they were a fix.

viskores/filter/contour/worklet/Clip.h
spyridon97: Well the explaination is. simple... do exactly what VTK does.

In my problematic commit i did everything below

```
                if (point1ToPoint2 < 0)
                {
                  viskores::Swap(ei.Vertex1, ei.Vertex2);
                  point1ToPoint2 = -point1ToPoint2;
                }
                auto point1ToIso =
                  0.0 - (static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - this->IsoValue));
                  0.0 - static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - this->IsoValue);
                ei.Weight = point1ToPoint2 != 0 ? point1ToIso / point1ToPoint2 : 0;
```

But i had forgot to check what we do above in the point1ToPoint2 calculation.

Besically the computation tries to imitate

https://github.com/Kitware/VTK/blob/master/Common/DataModel/vtkQuad.cxx#L778

but cell clipping does not require ascending edge vertices, so that's the only thing that is different which is why in 

PerformEdgeInterpolations

we swap back to make sure we have the correct weight. 

Also to answer your question

```
                auto point1ToIso =
                  0.0 - static_cast<viskores::Float64>(scalars.Get(ei.Vertex1) - this->IsoValue);
```

is mathematically equilavent to

```
                auto point1ToIso = this->IsoValue - static_cast<viskores::Float64>(scalars.Get(ei.Vertex1)
```

kmorel: OK. I get it. I saw the deleted parentheses and in my head I removed them from the expression, but they are still there in the `static_cast`. Still, I think this expression would be a lot more clear as

```cpp
viskores::Float64 point1toIso = this->IsoValue - scalars.Get(ei.Vertex1);
```

Anyway, the real fix is line 408 where the vertex values were flipped. Got it.

kmorel (approved):

spyridon97:
@kmorel i don't think i have write access to merge (unless you give it to me) so please merge.

PR #340, opened 2026-06-01
Properly check for ANARI cone extension

kmorel:
UnitTestANARIMapperGlyphs will skip the test if the KHR_GEOMETRY_CONE extension is missing (as it is in the Viskores device). However, it was always skipping (even for helide, which supports the extension) because it was checking the wrong string. The string to compare is fixed.


dpugmire (approved):
looks good.

PR #342, opened 2026-06-02
Replace ContourTreeUniformDistributed InternalStruct with local ExecutionState

ghweber:
Follow-up requested in https://github.com/Viskores/viskores/pull/316#discussion_r3326677244. Parent #316 is merged and this branch is rebased onto `main`, so the diff is just the `ExecutionState` refactor.

## Motivation

PR #316 introduced `std::shared_ptr<InternalStruct> Internals` to hold the `ContourTreeUniformDistributed` filter's intermediate results. Because filters can be copied, copies share this mutable execution scratch and can race or clobber each other when executed independently. @kmorel asked that the fix land in a separate PR with smaller scope — this is that PR.

## Change

- Rename `InternalStruct` → `ExecutionState` and **remove it from filter member state**.
- `ExecutionState` gains an `explicit` constructor that sizes the per-block result vectors to the number of partitions, so it is always correctly sized.
- Construct `ExecutionState` locally in `DoExecutePartitions` and pass it by reference through `ComputeLocalTree`, `ComputeLocalTreeImpl`, `PostExecute`, and `DoPostExecute`.
- `PreExecute` reverts to validation-only (no longer touches the result vectors), which also trims redundant `GetGlobalNumberOfPartitions()` calls.
- Move `ComputeLocalTree`/`ComputeLocalTreeImpl` to `private` (filter-internal; no external callers).
- The user-set configuration members (`BlocksPerDimension`, `LocalBlockIndices`) remain filter state — they are not execution scratch.

Copies of the filter no longer share execution scratch.

## CI

Builds and tests pass on macOS, Windows, and 17/18 GitLab configs (all CUDA / HIP+Kokkos / Clang / GCC / sanitizer / AlmaLinux). The single `ubuntu2004_gcc9` failure is an infrastructure error (container networking / out-of-memory during runner setup) that never reached compilation, and `ubuntu2004_gcc9_cuda118` with the same compiler passed.

### DIFF
--- a/viskores/filter/scalar_topology/ContourTreeUniformDistributed.cxx
+++ b/viskores/filter/scalar_topology/ContourTreeUniformDistributed.cxx
@@ -101,8 +101,22 @@ namespace filter
 namespace scalar_topology
 {
 
-struct ContourTreeUniformDistributed::InternalStruct
+struct ContourTreeUniformDistributed::ExecutionState
 {
+  /// Construct the execution state with all per-block result vectors sized to the
+  /// global number of data blocks being processed.
+  explicit ExecutionState(viskores::Id numberOfPartitions)
+    : LocalMeshes(static_cast<std::size_t>(numberOfPartitions))
+    , LocalContourTrees(static_cast<std::size_t>(numberOfPartitions))
+    , LocalBoundaryTrees(static_cast<std::size_t>(numberOfPartitions))
+    , LocalInteriorForests(static_cast<std::size_t>(numberOfPartitions))
+  {
+  }
+
+  /// The execution state is always passed by reference, never copied.
+  ExecutionState(const ExecutionState&) = delete;
+  ExecutionState& operator=(const ExecutionState&) = delete;
+
   /// Intermediate results (one per local data block)...
   /// ... local mesh information needed at end of fan out
   std::vector<viskores::worklet::contourtree_augmented::DataSetMesh> LocalMeshes;
@@ -117,7 +131,7 @@ struct ContourTreeUniformDistributed::InternalStruct
   //
   //std::vector<viskores::worklet::contourtree_distributed::HierarchicalContourTree> HierarchicalContourTrees;
   /// Number of iterations used to compute the contour tree
-  viskores::Id NumIterations;
+  viskores::Id NumIterations = 0;
 };
 
 //-----------------------------------------------------------------------------
@@ -201,7 +215,6 @@ ContourTreeUniformDistributed::ContourTreeUniformDistributed(
   viskores::cont::LogLevel treeLogLevel)
   : TimingsLogLevel(timingsLogLevel)
   , TreeLogLevel(treeLogLevel)
-  , Internals(new InternalStruct)
 {
   this->SetOutputFieldName("resultData");
 }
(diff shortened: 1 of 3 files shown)


ghweber:
@kmorel — the `Frank CI` pipeline shows one failed job, `ubuntu2004_gcc9` ([job log](https://gitlab.spack.io/viskores/viskores/-/jobs/22458729)), but it's an infrastructure failure, not a code issue:

```
ERROR: Job failed (system failure): Error response from daemon:
failed to set up container networking: ... cannot allocate memory
```

GitLab classifies it as a `system failure` — the runner couldn't allocate memory to set up container networking, so the job never reached compilation. Every other config passed, including `ubuntu2004_gcc9_cuda118` (same compiler), all other CUDA/HIP/Kokkos/Clang/GCC/sanitizer builds, plus macOS and Windows.

I don't have a gitlab.spack.io login to retry it. Could you click **Retry** on that one job when you get a chance? It should go green on a healthy runner. Thanks!

kmorel:
> I don't have a gitlab.spack.io login to retry it. Could you click **Retry** on that one job when you get a chance? It should go green on a healthy runner. Thanks!

I restarted the job. Let me know if I need to restart it again.

viskores/filter/scalar_topology/ContourTreeUniformDistributed.cxx
```diff
@@ -101,8 +101,18 @@ namespace filter
...
+  /// Construct the execution state with all per-block result vectors sized to the
+  /// global number of data blocks being processed.
+  explicit ExecutionState(viskores::Id numberOfPartitions)
+    : LocalMeshes(static_cast<std::size_t>(numberOfPartitions))
+    , LocalContourTrees(static_cast<std::size_t>(numberOfPartitions))
+    , LocalBoundaryTrees(static_cast<std::size_t>(numberOfPartitions))
+    , LocalInteriorForests(static_cast<std::size_t>(numberOfPartitions))
+  {
+  }
+
```
kmorel, line 115: I suggest disabling the copy constructor/operator for this class to prevent call-by-value. In addition to being less efficient, it could create subtle issues where changes in a called function do not propogate to the callee. If I remember my C++ properly, that would look like this.


```suggestion

  ExecutionState(const ExecutionState&) = delete;
  ExecutionState& operator=(const ExecutionState&) = delete;
```

kmorel (commented):
Just one minor suggested change.

viskores/filter/scalar_topology/ContourTreeUniformDistributed.cxx
ghweber: Good call — done in d1dadb2ee. I deleted the copy constructor and assignment operator (the state is always passed by reference), so call-by-value is now a compile error.

kmorel (approved):
Looks good once the CI passes.

It looks like the jobs were struggling last night. I restarted the failed jobs.

ghweber:
@kmorel — thanks for the approval and the restarts! Almost there: 17/19 Frank CI jobs are passing now. The only two still failing are the HIP/Kokkos builds:

- `ubuntu2204_hip_kokkos37`
- `ubuntu2204_hip_kokkos43`

Both failed with `stuck_pending_no_matching_runners` rather than a build error — they never landed on a runner, so it looks like there wasn't an available HIP runner to pick them up. Could you retry just those two once the HIP runners are back online? Should go green on a healthy runner. Thanks again!

kmorel:
> Both failed with `stuck_pending_no_matching_runners`

The appropriate runner(s) must have been down. I just restarted your failed jobs. One has already started, so that is a good sign.

ghweber:
Thanks. All tests have passed now, so I merged and closed the pull request.

PR #345, opened 2026-06-02
Add missing test and advertisement for ANARI sphere geometry.

dpugmire:
Advertise and validate the existing ANARI sphere geometry support. This adds khr_geometry_sphere to the device JSON and adds an image-based sphere geometry test/baseline so the advertised feature is covered by the same validation path as quad and cylinder.


dpugmire:
@kmorel, this is ready for a review now.

dpugmire:
@kmorel ping.

kmorel (approved):
Looks good. Thanks.

PR #347, opened 2026-06-03
Device adapter algorithms for user's guide

kmorel:

PR #349, opened 2026-06-07
Fix Reduce optimization in Kokkos adapter

kmorel:
Ran into a problem with optimization of calling CountBitSet with only one value in the array. It was caused by getting a value on the host when the portal was specialized to only work on the device.


kmorel:
@vicentebolea Please review. The testing for PR #347 revealed a bug with the Kokkos device adapter that could seg fault in rare circumstances (https://open.cdash.org/tests/2591109735). This PR should fix that.

vicentebolea (approved):

PR #350, opened 2026-06-09
Add SSIM Filter

kmorel:


kmorel:
@spyridon97 You suggested converting Viskores' image tests to use SSIM. Here is the first step, the introduction of an SSIM filter. Can you review to verify that this is on the right track?

@vicentebolea I could use a quick review from you as well.

vicentebolea (approved):
Looks good, I have not check in very detail but it looks alright and builds fine.

kmorel:
@spyridon97 Can you give this SSIM filter a quick review?

spyridon97:
---                                                                                                                              
  Reviewed against vtkImageSSIM as the reference. The SSIM formula itself is correct, and the Gaussian weighting is actually better than VTK (which computes weights but then
  hardcodes w = 1.0 on both passes). Two issues need addressing before this is mergeable.

  ---
  Bug: Multi-component statistics are combined instead of averaged per component
  
  vtkImageSSIM iterates for (int dim = 0; dim < nComp; ++dim) and computes mean/variance/covariance independently per component, writing a separate SSIM value per component per
  point. Viskores accumulates all components into a single SSIMStatistics struct via SSIMStatisticsAdd and finalizes once.

  This is mathematically wrong. Combining statistics introduces a cross-channel variance inflation: for two components with weighted means μ₀ and μ₁ the combined variance exceeds
  the average of the per-component variances by (μ₀ − μ₁)² / 4. The SSIM will be artificially low whenever channels have different means. For a single-component field the bug has
  no effect.
  
  The fix is to finalize inside the component loop and accumulate SSIM values (not statistics):

```cpp
  viskores::cont::ArrayHandle<viskores::Float64> output;
  for (viskores::IdComponent component = 0; component < numberOfComponents; ++component)
  {
      viskores::cont::ArrayHandle<SSIMStatistics> componentStatistics;
      invoke(SSIMComponentNeighborhood(patchRadius),
             inputCellSet,
             primary.GetComponentArray(component),                                                                                                               
             secondary.GetComponentArray(component),
             componentStatistics);                                                                                                 
                                                                                                                                                                 
      viskores::cont::ArrayHandle<viskores::Float64> componentSSIM;
      invoke(SSIMFinalize(dynamicRange, k1, k2), componentStatistics, componentSSIM);

      if (component == 0)
      {
          output = componentSSIM;                                                                                                  
      }                                                                                                                                                          
      else
      {
          invoke(SSIMStatisticsAdd{}, output, componentSSIM); // replace with element-wise Float64 add
      }
  }
  // divide output by numberOfComponents
```
  
  ---                                                                                                                                                            
  Numerical stability: one-pass variance vs. VTK's two-pass
  
  VTK computes variance in a dedicated second pass as Σ w(xᵢ − μ)². Viskores accumulates SumPrimarySquared and uses the one-pass identity E[X²] − E[X]².

  For standard image data (uint8 / float32, values well below 10⁶) this is fine because SSIMStatistics uses Float64 throughout. For scientific fields with large magnitudes and
  small local variation the cancellation loses significant bits. Since this filter targets scientific data beyond images, a two-pass approach or a compensated accumulator would be
  more robust. Not a blocker, but worth a follow-up.
  
  ---                                                                                                                                                            
  Note: Gaussian weights behave differently from VTK (intentional improvement)
  
  vtkImageSSIM computes Gaussian weights but overrides them with w = 1.0 on both the mean and variance passes, making it a flat-window SSIM in practice. Viskores correctly applies
  the Gaussian, which matches the Wang et al. 2004 paper more closely. This is a good divergence from VTK, but should be called out in the documentation so users are not surprised
  by the difference in results.

kmorel:
Thanks for the thorough analysis, @spyridon97. I'm glad I asked.

I believe I made the changes you suggested. Can you do one more pass to verify that it is correct now?

spyridon97:
I am on it.

kmorel:
@spyridon97 Ping.

spyridon97 (approved):
All good now!

PR #355, opened 2026-06-11
add cone to anari

dpugmire:
Add ANARI cone geometry support to the Viskores device.

This adds a cone geometry implementation, advertises `KHR_GEOMETRY_CONE`, wires cone creation into the ANARI device, and adds a direct ANARI cone geometry render test covering indexed and non-indexed cones, varying endpoint radii, caps, and primitive/vertex attributes.

### DIFF
--- a/viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
+++ b/viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
@@ -1,0 +1,520 @@
+//=============================================================================
+//
+//  The contents of this file are covered by the Viskores license. See
+//  LICENSE.txt for details.
+//
+//  By contributing to this file, all contributors agree to the Developer
+//  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+//
+//=============================================================================
+
+#include "Cone.h"
+// Viskores
+#include <viskores/CellShape.h>
+#include <viskores/VectorAnalysis.h>
+#include <viskores/cont/ArrayCopy.h>
+#include <viskores/cont/ArrayHandle.h>
+#include <viskores/cont/ArrayHandlePermutation.h>
+#include <viskores/cont/ArrayHandleRuntimeVec.h>
+#include <viskores/cont/CellSetSingleType.h>
+#include <viskores/rendering/CanvasRayTracer.h>
+#include <viskores/rendering/raytracing/RayOperations.h>
+#include <viskores/rendering/raytracing/RayTracer.h>
+#include <viskores/rendering/raytracing/TriangleExtractor.h>
+#include <viskores/rendering/raytracing/TriangleIntersector.h>
+// std
+#include <cmath>
+#include <vector>
+
+namespace viskores_device
+{
+namespace
+{
+
+constexpr viskores::IdComponent ConeSegments = 24;
+constexpr viskores::Float32 TwoPi = 6.2831853071795864769f;
+constexpr viskores::Float32 Epsilon = 1.e-6f;
+
+void AddTriangle(std::vector<viskores::Vec3f_32>& points,
+                 std::vector<viskores::Id>& connectivity,
+                 std::vector<viskores::Id>& primitiveIds,
+                 std::vector<viskores::Id>& vertexIds,
+                 viskores::Id primitiveId,
+                 const viskores::Vec3f_32& a,
+                 viskores::Id aVertexId,
+                 const viskores::Vec3f_32& b,
+                 viskores::Id bVertexId,
+                 const viskores::Vec3f_32& c,
+                 viskores::Id cVertexId)
+{
+  const viskores::Id firstPoint = static_cast<viskores::Id>(points.size());
+  points.push_back(a);
+  points.push_back(b);
+  points.push_back(c);
+  connectivity.push_back(firstPoint);
+  connectivity.push_back(firstPoint + 1);
+  connectivity.push_back(firstPoint + 2);
+  primitiveIds.push_back(primitiveId);
+  vertexIds.push_back(aVertexId);
+  vertexIds.push_back(bVertexId);
+  vertexIds.push_back(cVertexId);
+}
+
+struct AddExpandedField
+{
+  viskores::cont::DataSet* DataSet;
+  std::string FieldName;
+  viskores::cont::Field::Association Association;
+  viskores::cont::ArrayHandle<viskores::Id> SourceIds;
+
+  template <typename FieldArrayType>
+  void operator()(const FieldArrayType& fieldArray) const
+  {
+    viskores::cont::ArrayHandle<typename FieldArrayType::ValueType> expandedField;
+    viskores::cont::ArrayCopy(
+      viskores::cont::make_ArrayHandlePermutation(this->SourceIds, fieldArray), expandedField);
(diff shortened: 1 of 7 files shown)


viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
```diff
@@ -0,0 +1,520 @@
...
+#include <cmath>
+#include <vector>
+
+namespace viskores_device
+{
+namespace
+{
+
+constexpr viskores::IdComponent ConeSegments = 24;
+constexpr viskores::Float32 TwoPi = 6.2831853071795864769f;
```
kmorel, line 35: Viskores already provides the 2*pi constant (in both float32 and float64): https://viskores.readthedocs.io/en/latest/math.html#_CPPv4I0EN8viskores5TwoPiEN6detail23FloatingPointReturnTypeI1TE4TypeEv

viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+  template <typename FieldArrayType>
+  void operator()(const FieldArrayType& fieldArray) const
+  {
+    viskores::cont::ArrayHandle<typename FieldArrayType::ValueType> expandedField;
+    viskores::cont::ArrayCopy(
+      viskores::cont::make_ArrayHandlePermutation(this->SourceIds, fieldArray), expandedField);
+    this->DataSet->AddField(this->FieldName, this->Association, expandedField);
+  }
+};
```
kmorel, line 78: You shouldn't need this struct. Use `viskores::filter::MapFieldPermutation` to permute an `UnknownArrayHandle` or `Field` object.

https://viskores.readthedocs.io/en/latest/extended-filter-impl.html#_CPPv4N8viskores6filter19MapFieldPermutationERKN8viskores4cont5FieldERKN8viskores4cont11ArrayHandleIN8viskores2IdEEERN8viskores4cont5FieldEN8viskores7Float64E

viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
```diff
@@ -0,0 +1,520 @@
...
+    case 2:
+      return "attribute1";
+    case 3:
+      return "attribute2";
+    case 4:
+      return "attribute3";
+    default:
+      return "";
+  }
+}
```
kmorel, line 97: I think it will be easier to delete this and iterate over a static `array` with these attributes. (See comment around line 403.)

viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
```diff
@@ -0,0 +1,520 @@
...
+               viskores::CELL_SHAPE_TRIANGLE,
+               3,
+               connectivityArray);
+  this->m_dataSet.SetCellSet(cellSet);
+  this->m_dataSet.AddCoordinateSystem("coords", pointsArray);
+
+  for (viskores::IdComponent attribute = 0; attribute < 5; ++attribute)
+    this->AddExpandedPrimitiveField(AttributeName(attribute), primitiveIdArray, numberOfCones);
+  for (viskores::IdComponent attribute = 0; attribute < 5; ++attribute)
+    this->AddExpandedVertexField(AttributeName(attribute), vertexIdArray, numberOfVertices);
```
kmorel, line 406: Matching the comment above, it seems easier to define an array with the appropriate strings and iterate over that.


```suggestion
  static const std::array<std::string, 5> attributeNames = {
    "color", "attribute0", "attribute1", "attribute2", "attribute3"
  };
  for (auto&& attribute : attributeNames)
  {
    this->AddExpandedPrimitiveField(attribute, primitiveIdArray, numberOfCones);
    this->AddExpandedVertexField(attribute, vertexIdArray, numberOfVertices);
  }
```

viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
```diff
@@ -0,0 +1,520 @@
...
+  {
+    viskores::UInt8 capMask = 0;
+    if (caps.Get(2 * coneId) != 0)
+      capMask |= 1;
+    if (caps.Get(2 * coneId + 1) != 0)
+      capMask |= 2;
+    return capMask;
+  }
+  if (numberOfCaps >= numberOfCones && coneId < numberOfCaps)
+    return caps.Get(coneId) != 0 ? viskores::UInt8(3) : viskores::UInt8(0);
```
kmorel, line 147: According to the [ANARI specification](https://registry.khronos.org/ANARI/specs/1.1/ANARI-1.1.html#object_types_geometry_cone), the `vertex.color` parameters is "per-vertex end cap flags". I don't see the `caps` array coming from anywhere else. So where do these conditions of `numberOfCones` come from?

viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
```diff
@@ -0,0 +1,520 @@
...
+    viskores::cont::ArrayCopyShallowIfPossible(this->m_vertexRadius->dataAsViskoresArray(), radii);
+  }
+  else
+  {
+    radii.AllocateAndFill(positions.GetNumberOfValues(), this->m_globalRadius);
+  }
+
+  viskores::cont::ArrayHandle<viskores::UInt8> caps;
+  if (this->m_vertexCaps)
+    viskores::cont::ArrayCopyShallowIfPossible(this->m_vertexCaps->dataAsViskoresArray(), caps);
```
kmorel, line 204: The implementation will be simpler if `caps` were filled with the proper value if not specified. (See similar comment in `GetCapMask` method.


```suggestion
  viskores::cont::ArrayHandle<viskores::UInt8> caps;
  if (this->m_vertexCaps)
  {
    viskores::cont::ArrayCopyShallowIfPossible(this->m_vertexCaps->dataAsViskoresArray(), caps);
  }
  else
  {
    caps.AllocateAndFill(positions.GetNumberOfValues(), this->m_globalCapMask);
  }
```

viskores/rendering/anari-device/scene/surface/geometry/Cone.cpp
```diff
@@ -0,0 +1,520 @@
...
+  std::vector<viskores::Vec3f_32> meshPoints;
+  std::vector<viskores::Id> connectivity;
+  std::vector<viskores::Id> primitiveIds;
+  std::vector<viskores::Id> vertexIds;
+  meshPoints.reserve(static_cast<std::size_t>(numberOfCones * ConeSegments * 12));
+  connectivity.reserve(static_cast<std::size_t>(numberOfCones * ConeSegments * 12));
+  primitiveIds.reserve(static_cast<std::size_t>(numberOfCones * ConeSegments * 4));
+  vertexIds.reserve(static_cast<std::size_t>(numberOfCones * ConeSegments * 12));
+
+  for (viskores::Id coneId = 0; coneId < numberOfCones; ++coneId)
```
kmorel, line 232: Having a for loop over the number of cones seems like a bad idea. This can be quite large, and it would make more sense to build the geometry in a worklet.

It makes more sense to implement this functionality in its own filter. I suggest putting this PR aside, implementing a `Glyph` filter in a separate PR, and then come back to PR and change the implementation to use this filter to generate the cone geometry.

kmorel (changes requested):

PR #361, opened 2026-06-23
Refactor ContourTreeAugmented: DataSet output, AugmentationType enum, ProcessContourTree helpers

ghweber:
## Summary

Refactors `ContourTreeAugmented` to return its results via the output `DataSet` and modernizes its API:

- **Output `DataSet` fields** (single-`DataSet` execution): `Supernodes` and `Superarcs` are always written; `Superparents` (a point field mapping each mesh vertex to its superarc) is added when the tree is augmented; and `WhichBranch`, `BranchMinimum`, `BranchMaximum`, `BranchSaddle`, `BranchParent` are added when branch decomposition is enabled. All fields are indexed by mesh vertex or by supernode directly -- no separate sort order is needed to interpret them. The field names are a fixed contract defined in `ContourTreeAugmentedFieldNames.h`.
- **Optional `SortOrder` export**: `SetIncludeSortOrder(true)` additionally writes the `SortOrder` field (the sorted-to-mesh-index permutation the tree was computed with), so downstream algorithms that operate in sorted space can reconstruct it with a single scatter instead of re-sorting. See "Design discussion" below for why this is the chosen mechanism.
- **Boolean augmentation API**: `SetAugmentTree(bool)` / `GetAugmentTree()` replace the constructor's `computeRegularStructure` argument; the old constructor is kept as deprecated. (Boundary augmentation, level 2, remains reachable only through the deprecated constructor for the deprecated multi-block path; no new setter for it is introduced.)
- **`SetComputeBranchDecomposition(true)`** enforces a fully augmented tree at `Execute()` time (logging a warning if it overrides the current setting), so setter order does not matter.
- **`GetContourTreeStatistics()`** returns a `ContourTreeStatistics` object (array sizes plus per-iteration hyperstructure statistics) with individual members and a `ToString()` / `operator<<` conversion for logging.
- **Multi-block (`PartitionedDataSet`/MPI) path deprecated**: computing a monolithic contour tree for the entire data set is inherently inefficient in parallel -- boundary augmentation was only a way to make it somewhat efficient -- and the path was largely kept for performance comparisons before a full distributed analysis pipeline existed. Use `ContourTreeUniformDistributed` together with `DistributedBranchDecompositionFilter`, `SelectTopVolumeBranchesDistributedFilter`, and `ExtractTopVolumeContoursFilter` instead. The deprecated path keeps its established behavior (results reachable via `GetContourTree()`, `GetSortOrder()`, and `GetNumIterations()`).
- **DataSet-based helpers in `ProcessContourTree`**: `CollectSortedSuperarcs`, `CollectRegularVerticesPerSuperarc`, `SelectTopVolumeBranches`, and mesh-vertex-space `ComputeBranchDecomposition` overloads operate directly on the filter's output `DataSet` -- callers need neither sort order nor raw internal arrays. Mesh vertex space is the default for these helpers; the sortOrder-taking `ComputeBranchDecomposition` overload remains only for the deprecated path. The two collection helpers are computed with Viskores device primitives (the new `SuperarcToEdge` worklet plus `viskores::cont::Algorithm`), not host loops. `CollectRegularVerticesPerSuperarc` returns a grouped array (flat values + per-superarc offsets); a convenience overload plus the `GroupedArrayToNestedVectors` helper materialize it as a `std::vector<std::vector<viskores::Id>>` for host-side consumers. Pre-existing serial helpers are unchanged.
- **Rename `SelectTopVolumeBranchesFilter` -> `SelectTopVolumeBranchesDistributedFilter`** to reflect that it belongs to the distributed pipeline; the old header keeps a deprecated `using` alias. This also prepares for a future PR that splits the non-distributed branch decomposition computation out of `ContourTreeAugmented` into a separate filter.
- **Deprecated-path tests isolated**: tests for deprecated functionality (boundary augmentation, multi-block path with legacy getters) live in a new `UnitTestContourTreeUniformAugmentedFilterDeprecated.cxx` so they can be deleted together with the deprecated code; the main unit test no longer references any deprecated API.
- **`ContourTreeApp` example** updated to the new API (the deprecated multi-block path is retained under `WITH_MPI`, using the deprecated constructor).
- **New `contour_tree_augmented_simple` example**: demonstrates filter usage and output-field access, prints the superarc segmentation of the mesh, and writes the contour tree as a GraphViz `.dot` file. (Branch decomposition is deliberately not shown here; a dedicated example will accompany the planned branch-decomposition filter.)

### Background: the two index spaces

The contour tree algorithm works internally in *sorted space*: vertices are identified by their rank in the (value, mesh index) lexicographic order (`SimulatedSimplicityIndexComparator`), which encodes simulated simplicity -- ties in the scalar value are broken by mesh index. The previous API exposed results in this space: `Supernodes` contained sort IDs, and callers needed `GetSortOrder()` (the sortId -> meshId permutation) to relate any result to the input mesh. We assume the filter is easier to use when results are translated to mesh index space, where they can be related to the input mesh directly; sorted-space results also do not fit the `DataSet` model -- a per-vertex array in sorted order cannot carry the Points association, because point fields are by definition indexed in mesh vertex order.

Sorted space is not just an implementation detail, though; it provides real capabilities:

1. *Value comparisons as index comparisons.* `sortId(a) < sortId(b)` is exactly "a below b" under simulated simplicity, so plain `viskores::Minimum`/`Maximum` over sort IDs implement value ordering -- the `ProcessContourTree` hypersweeps rely on this.
2. *O(1) rank queries and global extrema.* A vertex's sort ID **is** its rank; the global minimum/maximum are simply positions 0 and n-1 (the height branch decomposition uses `Superparents[0]` / `Superparents[n-1]` this way).
3. *Iteration in value order* by walking the sort order array.
4. *The exact tie-broken order the tree was built with*, guaranteed consistent with every decision made during construction.

The key observation that makes the mesh-space output safe: because the sort order is fully determined by (value, mesh index), every sorted-space comparison can be reproduced in mesh space by comparing (value[i], i) pairs -- nothing is *destroyed* by the transformation, only the cached rank information is discarded. And the expensive part of recovering it is the comparison sort, not the permutation: given the `SortOrder` array, the inverse map (meshId -> sortId) is one O(n) scatter, after which any array moves between spaces with cheap gathers.

### The decision tree

**Decision 1: should the filter transform its output to mesh index space at all?** Yes. Self-describing output (fields indexed by mesh vertex or supernode, `Superparents` as a genuine point field usable for mesh segmentation) is the main point of this refactor, and per the analysis above the transformation loses no capability that cannot be recovered. The alternative -- keep sorted-space output and a `SortOrder` field as the primary contract -- would require every consumer to deal with two index spaces to save a gather for the minority of consumers that want sorted space.

**Decision 2: how do downstream algorithms that want sorted space (notably a future branch-decomposition filter) get it back?** We considered four options:

**Option 0 -- recompute the sort order downstream when needed.**
*Pros*: no filter API surface at all; the sort order is recoverable from the field values plus mesh IDs alone, since the comparator is deterministic.
*Cons*: pays a full O(n log n) device comparison sort again -- the sort is a significant fraction of contour tree construction time on large data, and repeating work the filter already did is exactly what an output contract should avoid. It also relies on the downstream comparator staying bit-for-bit consistent with the filter's; a drift in tie-breaking would silently produce a *different* valid-looking sorted space. This option remains available as a fallback (nothing is lost without the flag), but it should not be the designed path.

**Option A -- a filter flag that switches the space of the output arrays**, with space-specific field names (e.g. `SupernodesMeshId` vs. `SupernodesSortId`), plus `SortOrder` in the sorted case.
*Pros*: sorted-space consumers get their arrays with zero conversion; consumers state explicitly which space they want, and wrong-space access fails fast on the field name.
*Cons*: the output contract becomes conditional on a flag -- the same conceptual result has two shapes. `Superparents` cannot be a point field in the sorted-space variant (point fields are mesh-indexed by definition), so the two variants differ even in field associations. Every `ProcessContourTree` DataSet helper must either handle both spaces or reject one at runtime. Two naming schemes must be documented and maintained, and pipelines break when someone flips the flag.

**Option B (chosen) -- one canonical space (mesh index) for all output arrays, plus an opt-in `SortOrder` field** (`SetIncludeSortOrder(true)`).
*Pros*: every output field has a single, unconditional meaning, and the only space-specific name in the contract is `SortOrder` itself. Point fields keep their natural Points association. A downstream filter inverts `SortOrder` with one O(n) scatter and can then gather any array back into sorted space -- no re-sort ever happens, and the existing sorted-space algorithms (the `ProcessContourTree` hypersweeps, min/max sweeps with plain `Minimum`/`Maximum` operators) can be reused unchanged behind a small conversion prologue. Exporting the permutation preserves the exact tie-broken order the tree was built with instead of trusting a downstream comparator to reproduce it (the weakness of Option 0). The capabilities enumerated above all come back: rank queries via the inverted permutation, global extrema at positions 0 / n-1 of `SortOrder`, value-order iteration by walking it.
*Cons*: sorted-space consumers pay one scatter plus gathers (O(n) memory-bound operations, negligible next to tree construction); the flag must be set before `Execute()` if the permutation is needed later (else fall back to Option 0's re-sort); one more filter option to document.

**Option C -- always export `SortOrder`.**
*Pros*: simplest contract, nothing to remember, sorted space always recoverable.
*Cons*: adds an array with one entry per mesh vertex to every result. Most analysis consumers (isovalue selection, segmentation, branch queries) never touch it, so for large data this is a pure memory tax on the common case.

### Rationale for Option B

Option B is the only option that simultaneously (a) keeps a single, unconditional output contract that fits the `DataSet` field model, (b) loses no capability -- everything expressible in sorted space is reconstructible in O(n) from the mesh-space output plus `SortOrder`, and the sort order itself remains recoverable by re-sorting even without the flag, (c) avoids the memory tax on consumers that don't need sorted space, and (d) keeps the planned split of the branch decomposition into a separate filter cheap to implement, since that filter can request `SortOrder` and run the existing hypersweep code unchanged. (For that future filter, the hyperstructure arrays -- `Hypernodes`, `Hyperarcs`, `Hyperparents`, `WhenTransferred`, plus `NumIterations` -- are index-space-neutral and can be exported the same way when the time comes; that export is deliberately left to the future PR where its exact shape is decided.)

## Test plan

- [ ] `UnitTestContourTreeUniformAugmentedFilter` and `UnitTestContourTreeUniformAugmentedWorklet` pass, including new tests: branch decomposition forces full augmentation regardless of setter order; `CollectRegularVerticesPerSuperarc` / `GroupedArrayToNestedVectors` are verified against the `Supernodes`/`Superparents` fields they are derived from; `SortOrder` is exported only on request and is the (value, mesh index) permutation
- [ ] `UnitTestContourTreeUniformAugmentedFilterDeprecated` passes (boundary augmentation matches the supported API; multi-block path with legacy getters)
- [ ] `ContourTree_Augmented` and `ContourTreeAugmentedSimple` examples build and run correctly
- [ ] `SourceInInstall` passes
- [ ] Frank CI

https://claude.ai/code/session_01UbGvfaJxTf8XHNebsZWgVw


ghweber:
Hamish Carr reviewed this PR via email and provided the following feedback:

> OK. I think I understand what is going on.
>
> I haven't checked the logic in detail, but I think the changes are mostly reasonable.
>
> There's a typo where you have `branchDecompostionRoot` instead of `branchDecompositionRoot`: you might want to fix that.
>
> H

The typo has been fixed in the latest commit. Hamish was unable to post directly due to a GitHub account issue.

docs/changelog/refactor-contour-tree-augmented.md
```diff
@@ -0,0 +1,41 @@
+## ContourTreeAugmented returns results via the output DataSet; deprecates the multi-block path
+
+`ContourTreeAugmented` now returns its results as fields on the output
+`DataSet` when run on a single `DataSet`: `Supernodes` and `Superarcs`
+(whole-dataset fields), and — when the tree is augmented — `Superparents` (a
```
ghweber: Avoid non ASCII characters like "—" in changelog or markdown files.

docs/changelog/refactor-contour-tree-augmented.md
```diff
@@ -0,0 +1,41 @@
...
+The augmentation level is now a simple boolean: use `SetAugmentTree(bool)` /
+`GetAugmentTree()` in place of the constructor's `computeRegularStructure`
+argument and the deprecated `SetComputeRegularStructure`/
+`GetComputeRegularStructure`. `SetComputeBranchDecomposition(true)` requires
+an augmented tree; if augmentation isn't already enabled, `Execute()` enables
+it automatically and logs a warning, regardless of the order you called the
+setters in.
+
+The multi-block (`PartitionedDataSet`/MPI) path of `ContourTreeAugmented` is
+now deprecated. Full augmentation of a contour tree computed in parallel is
```
ghweber: It's not just full augmentation that is inefficient to do in parallel. Just computing a monolithic tree for the entire data set is inefficient. That is why we switched to the distributed tree representation. Boundary augmentation was just a way to be able to compute the full tree at least somewhat efficiently. But even computing the unaugmented tree is better using the distributed version.

To some degree we implemented this for performance comparison and since we did not have a full analysis pipeline for the distributed case. The distributed case is still not a 1:1 swap in but new applications should extend that rather than build on a version that scales terribly.

examples/contour_tree_augmented/ContourTreeApp.cxx
```diff
@@ -687,18 +691,36 @@ int main(int argc, char* argv[])
...
-  viskores::filter::scalar_topology::ContourTreeAugmented filter(useMarchingCubes,
-                                                                 computeRegularStructure);
-
+  viskores::filter::scalar_topology::ContourTreeAugmented filter;
+  filter.SetUseMarchingCubes(useMarchingCubes);
 #ifdef WITH_MPI
+  // Deprecated multi-block path: use the deprecated tri-state augmentation knob (boundary
+  // augmentation is reachable only here) and drive branch decomposition outside the filter.
+  VISKORES_DEPRECATED_SUPPRESS_BEGIN
+  filter.SetComputeRegularStructure(computeRegularStructure);
```
ghweber: I am wondering if we are introducing `SetComputeRegularStructure` as we are deprecating it. If it wasn't already present, we should not add it and perhaps keep using the deprecated non-default constructor for the `WITH_MPI` case.

examples/contour_tree_augmented/ContourTreeApp.cxx
```diff
@@ -687,18 +691,36 @@ int main(int argc, char* argv[])
...
   filter.SetBlockIndices(blocksPerDim, localBlockIndices);
+  VISKORES_DEPRECATED_SUPPRESS_END
+#else
+  if (computeRegularStructure == 2)
+  {
+    VISKORES_LOG_S(viskores::cont::LogLevel::Warn,
+                   "Boundary augmentation (--augmentTree=2) is only supported by the deprecated"
+                   " multi-block (MPI) path. Computing the tree without augmentation.");
+  }
+  filter.SetAugmentTree(computeRegularStructure == 1);
```
ghweber, line 736: Longer term we should change the command line flag to be just a boolean augment flag.

examples/contour_tree_augmented/ContourTreeApp.cxx
```diff
@@ -687,18 +691,36 @@ int main(int argc, char* argv[])
...
 #endif
   filter.SetActiveField("values");
 
   // Execute the contour tree analysis. NOTE: If MPI is used the result  will be
   // a viskores::cont::PartitionedDataSet instead of a viskores::cont::DataSet
   auto result = filter.Execute(useDataSet);
 
+#ifndef WITH_MPI
+  // Single-DataSet path: results are read from the output DataSet fields.
+  const viskores::cont::DataSet& resultDS = result;
```
ghweber: Why create an alias here? Why not just use result later on in the code?

examples/contour_tree_augmented/ContourTreeApp.cxx
```diff
@@ -843,6 +864,62 @@ int main(int argc, char* argv[])
...
     } //end if compute isovalue
   }
+  VISKORES_DEPRECATED_SUPPRESS_END
+#else
+  ////////////////////////////////////////////
+  // Isovalue selection from branch decomposition
+  // (branch decomposition arrays were computed inside the filter when
+  //  SetComputeBranchDecomposition(true) was set)
+  ////////////////////////////////////////////
+  if (rank == 0 && computeBranchDecomposition && numLevels > 0)
```
ghweber: Isn't rank always 0 in the non-MPI version? This guards doesn't really make sense to me here.

ghweber: Otherwise, is there a better way to consolidate non-MPI and MPI version? The goal would be to make the minimum amount of changes needed to the code.

examples/contour_tree_augmented/ContourTreeApp.cxx
```diff
@@ -901,6 +987,13 @@ int main(int argc, char* argv[])
   VISKORES_LOG_S(viskores::cont::LogLevel::Info,
                  std::endl
                    << ct.PrintHyperStructureStatistics(false) << std::endl);
+#else
+  VISKORES_LOG_S(viskores::cont::LogLevel::Info,
+                 std::endl
+                   << "    ---------------- Contour Tree Statistics ---------------------"
+                   << std::endl
+                   << filter.GetContourTreeStatisticsString());
```
ghweber: Rather than returning a string here, perhaps return statistics as an object with individual members and that has conversion to string?

examples/contour_tree_augmented_simple/ContourTreeSimple.cxx
```diff
@@ -0,0 +1,241 @@
+//============================================================================
```
ghweber, line 1: Can we further simplify this example?

examples/contour_tree_distributed/ContourTreeApp.cxx
```diff
@@ -83,7 +83,7 @@
 #include <viskores/filter/scalar_topology/ContourTreeUniformDistributed.h>
 #include <viskores/filter/scalar_topology/DistributedBranchDecompositionFilter.h>
 #include <viskores/filter/scalar_topology/ExtractTopVolumeContoursFilter.h>
-#include <viskores/filter/scalar_topology/SelectTopVolumeBranchesFilter.h>
+#include <viskores/filter/scalar_topology/SelectTopVolumeBranchesDistributedFilter.h>
```
ghweber, line 86: Say somewhere that this is also in preparation for a future PR that splits the non-distributed branch decomposition computation into a separate filter.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -244,32 +246,55 @@ class TestContourTreeUniformAugmented
...
+                      std::vector<DataValueType>& isoValues) const
+  {
+    namespace caugmented_ns = viskores::worklet::contourtree_augmented;
+
     DataValueType eps = 0.00001f;      // Distance away from critical point
     viskores::Id numComp = levels + 1; // Number of components the tree should be simplified to
     bool usePersistenceSorter = true;
 
-    // Compute the branch decomposition
+    VISKORES_DEPRECATED_SUPPRESS_BEGIN
```
ghweber: Are any other filters suppress deprecation warnings? I would skip this. I would also consider moving all deprecated tests into one location; possibly a separate test file; to make it easier to remove them down the road.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -244,32 +246,55 @@ class TestContourTreeUniformAugmented
...
     caugmented_ns::IdArrayType hyperarcDependentWeight;
 
-    caugmented_ns::ProcessContourTree::ComputeVolumeWeightsSerial(
-      filter.GetContourTree(),
-      filter.GetNumIterations(),
-      superarcIntrinsicWeight,  // (output)
-      superarcDependentWeight,  // (output)
-      supernodeTransferWeight,  // (output)
-      hyperarcDependentWeight); // (output)
+    caugmented_ns::ProcessContourTree::ComputeVolumeWeightsSerial(filter.GetContourTree(),
```
ghweber: Why remove the // output comments; they are helpful

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -278,28 +303,18 @@ class TestContourTreeUniformAugmented
...
     caugmented_ns::ProcessContourTree::ComputeVolumeBranchDecompositionSerial(
       filter.GetContourTree(),
       superarcDependentWeight,
       superarcIntrinsicWeight,
-      whichBranch,   // (output)
-      branchMinimum, // (output)
-      branchMaximum, // (output)
-      branchSaddle,  // (output)
-      branchParent); // (output)
+      whichBranch,
```
ghweber: Why remove the // output comments; they are helpful

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -278,28 +303,18 @@ class TestContourTreeUniformAugmented
...
+      branchMinimum,
+      branchMaximum,
+      branchSaddle,
+      branchParent);
```
ghweber: Why are you removing comments. That is BAD.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -317,18 +332,14 @@ class TestContourTreeUniformAugmented
         filter.GetSortOrder(),
         dataField,
         dataFieldIsSorted);
+    VISKORES_DEPRECATED_SUPPRESS_END
 
-    // Simplify the contour tree of the branch decompostion
```
ghweber, line 321: Do not remove comments.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -337,17 +348,21 @@ class TestContourTreeUniformAugmented
...
-    unsigned int computeRegularStructure,
-    unsigned int dataSetNo) const
+  //
+  // When deprecatedBoundary is true the (now deprecated) boundary-augmentation level (2) is
+  // requested via the deprecated SetComputeRegularStructure API; otherwise augmentTree selects
+  // full (true) or no (false) augmentation through the supported bool API.
+  viskores::cont::DataSet RunContourTree(bool useMarchingCubes,
+                                         bool augmentTree,
+                                         unsigned int dataSetNo,
+                                         bool deprecatedBoundary = false) const
```
ghweber: Having two flags here `augmentTree` and `deprecatedBoundary` is confusing. Also `deprecatedBoundary` is not a good name.

ghweber: Possibly have a separate function instead that can later be deleted.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -366,32 +381,40 @@ class TestContourTreeUniformAugmented
...
   }
 
 public:
   //
   // Create a uniform 2D structured cell set as input with values for contours
   //
-  void TestContourTree_Mesh2D_Freudenthal_SquareExtents(
-    unsigned int computeRegularStructure = 1) const
+  void TestContourTree_Mesh2D_Freudenthal_SquareExtents(bool augmentTree = true,
+                                                        bool deprecatedBoundary = false) const
```
ghweber: Same comment about `deprecatedBoundary` as above

ghweber: Do we actually have any tests that run with boundary augmentation? If not, no need to add tests for something deprecated.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -424,21 +447,20 @@ class TestContourTreeUniformAugmented
                          "Wrong result for ContourTree filter");
   }
 
-  void TestContourTree_Mesh2D_Freudenthal_NonSquareExtents(
-    unsigned int computeRegularStructure = 1) const
+  void TestContourTree_Mesh2D_Freudenthal_NonSquareExtents(bool augmentTree = true,
+                                                           bool deprecatedBoundary = false) const
```
ghweber: See above.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -474,23 +496,22 @@ class TestContourTreeUniformAugmented
                          "Wrong result for ContourTree filter");
   }
 
-  void TestContourTree_Mesh3D_Freudenthal_CubicExtents(
-    unsigned int computeRegularStructure = 1) const
+  void TestContourTree_Mesh3D_Freudenthal_CubicExtents(bool augmentTree = true,
+                                                       bool deprecatedBoundary = false) const
```
ghweber: See above.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -530,23 +551,22 @@ class TestContourTreeUniformAugmented
                          "Wrong result for ContourTree filter");
   }
 
-  void TestContourTree_Mesh3D_Freudenthal_NonCubicExtents(
-    unsigned int computeRegularStructure = 1) const
+  void TestContourTree_Mesh3D_Freudenthal_NonCubicExtents(bool augmentTree = true,
+                                                          bool deprecatedBoundary = false) const
```
ghweber: See above.

viskores/filter/scalar_topology/testing/UnitTestContourTreeUniformAugmentedFilter.cxx
```diff
@@ -718,13 +849,16 @@ class TestContourTreeUniformAugmented
...
 
     std::string fieldName = "pointvar";
-    bool useMarchingCubes = false;
-    bool computeRegularStructure = true;
-    viskores::filter::scalar_topology::ContourTreeAugmented filter(useMarchingCubes,
-                                                                   computeRegularStructure);
-    filter.SetActiveField(fieldName);
+    std::vector<ValueType> isoValues;
 
 #ifdef VISKORES_ENABLE_MPI
```
ghweber: There should be a comment that these tests are for deprecated functions.

viskores/filter/scalar_topology/worklet/contourtree_augmented/processcontourtree/Branch.h
```diff
@@ -260,6 +277,37 @@ Branch<T>* Branch<T>::ComputeBranchDecomposition(
...
+  const IdArrayType& branchMaximum,
+  const IdArrayType& branchSaddle,
+  const IdArrayType& branchParent,
+  const viskores::cont::ArrayHandle<T, StorageType>& dataField)
+{ // ComputeBranchDecompositionMeshSpace()
+  // contourTreeSupernodes already contains mesh vertex IDs (not sorted indices) and dataField
+  // is indexed by mesh vertex, i.e., exactly the sorted-space inputs to ComputeBranchDecomposition
+  // with an identity sortOrder and dataFieldIsSorted=true (mesh index == "sorted" index here).
+  IdArrayType identitySortOrder;
+  viskores::cont::Algorithm::Copy(viskores::cont::ArrayHandleIndex(dataField.GetNumberOfValues()),
```
ghweber: Can we avoid this copy. Also, default should be to operate on contourTreeSupernodes that already contain mesh vertex IDs as the other is only accessible via deprecated functions.

viskores/filter/scalar_topology/worklet/contourtree_augmented/ProcessContourTree.h
```diff
@@ -751,6 +869,132 @@ class ProcessContourTree
       dataFieldIsSorted);
   }
 
+  // Create branch decomposition when arrays are already in mesh vertex space (no sortOrder needed).
+  // Use this when Supernodes contains mesh vertex IDs and Superparents is mesh-indexed —
+  // i.e., with the output DataSet fields from ContourTreeAugmented.
+  template <typename T, typename StorageType>
+  static process_contourtree_inc_ns::Branch<T>* ComputeBranchDecompositionMeshSpace(
```
ghweber: MeshSpace should be the default.

examples/contour_tree_augmented/ContourTreeApp.cxx
ghweber: Tracked in #380: the flag becomes a plain boolean when the deprecated multi-block path is removed.

mingzhefluorite:
A testing post for comment submission.

PR #364, opened 2026-06-27
fix(anari): return device extensions from getDeviceExtensions

marcelkrueger:
The ANARI device returned a nullptr for `getDeviceExtensions` which meant that any client (e.g. Paraview, but also ANARI-SDK examples) checking features via `anariGetDeviceExtensions` failed.

Example from calling `anariInfo` from ANARI-SDK:

Before:
```bash
➜  bin ./anariInfo -l viskores | grep -A 10 Extensions
[DEBUG] destroying Viskores device (0x10445ea20)
```

After fix:
```bash
➜  bin ./anariInfo -l viskores | grep -A 10 Extensions
[DEBUG] destroying Viskores device (0x10557ea20)
   Extensions:
      ANARI_KHR_CAMERA_ORTHOGRAPHIC
      ANARI_KHR_CAMERA_PERSPECTIVE
      ANARI_KHR_GEOMETRY_CYLINDER
      ANARI_KHR_GEOMETRY_QUAD
      ANARI_KHR_GEOMETRY_SPHERE
      ANARI_KHR_GEOMETRY_TRIANGLE
      ANARI_KHR_MATERIAL_MATTE
      ANARI_KHR_SAMPLER_IMAGE1D
      ANARI_KHR_VOLUME_TRANSFER_FUNCTION1D
   Subtypes:

```

This was previously not caught because the unit tests called 
```cpp
anariGetProperty(d, d, "extension", ANARI_STRING_LIST, &extensions, sizeof(char**), ANARI_WAIT);
```
instead of 
```cpp
ANARI_INTERFACE const char ** anariGetDeviceExtensions(ANARILibrary library, const char* deviceSubtype);
```

So it might make sense to also change the unit tests to use the `anariGetDeviceExtensions`.

If you agree I can add that as well before merge.

Best
Marcel

### DIFF
--- a/docs/changelog/anari-getdeciveextensions-fix.md
+++ b/docs/changelog/anari-getdeciveextensions-fix.md
@@ -1,0 +1,30 @@
+## Fix ANARI ViskoresLibrary getDeviceExtensions to return device extensions
+The ANARI device returned a nullptr for `getDeviceExtensions` which meant 
+that any client (e.g. Paraview, but also ANARI-SDK examples) checking 
+features via `anariGetDeviceExtensions` failed.
+
+Example from calling `anariInfo` from ANARI-SDK:
+
+Before:
+```bash
+➜  bin ./anariInfo -l viskores | grep -A 10 Extensions
+[DEBUG] destroying Viskores device (0x10445ea20)
+```
+
+After fix:
+```bash
+➜  bin ./anariInfo -l viskores | grep -A 10 Extensions
+[DEBUG] destroying Viskores device (0x10557ea20)
+   Extensions:
+      ANARI_KHR_CAMERA_ORTHOGRAPHIC
+      ANARI_KHR_CAMERA_PERSPECTIVE
+      ANARI_KHR_GEOMETRY_CYLINDER
+      ANARI_KHR_GEOMETRY_QUAD
+      ANARI_KHR_GEOMETRY_SPHERE
+      ANARI_KHR_GEOMETRY_TRIANGLE
+      ANARI_KHR_MATERIAL_MATTE
+      ANARI_KHR_SAMPLER_IMAGE1D
+      ANARI_KHR_VOLUME_TRANSFER_FUNCTION1D
+   Subtypes:
+
+```
--- a/viskores/rendering/anari-device/ViskoresLibrary.cpp
+++ b/viskores/rendering/anari-device/ViskoresLibrary.cpp
@@ -11,6 +11,7 @@
 #include "ViskoresDevice.h"
 #include "anari/backend/LibraryImpl.h"
 #include "anari_library_viskores_export.h"
+#include "anari_library_viskores_queries.h"
 
 #include <viskores/cont/Initialize.h>
 
@@ -45,7 +46,7 @@ ANARIDevice ViskoresLibrary::newDevice(const char* /*subtype*/)
 
 const char** ViskoresLibrary::getDeviceExtensions(const char* /*deviceType*/)
 {
-  return nullptr;
+  return query_extensions();
 }
 
 } // namespace viskores_device


kmorel (approved):
Thanks for the fix. This looks good.

If you want to update the tests to use `getDeviceExtensions`, that would be good, too.

marcelkrueger:
It needed a little more refactoring because the library wasn't exposed when loading with `loadANARIDevice` as far as I can see.
Therefore I am loading the lib similar to `loadANARIDevice` would, check extensions and unload in the helper function.

I simplified the unit tests accordingly.

kmorel:
> It needed a little more refactoring because the library wasn't exposed when loading with `loadANARIDevice` as far as I can see.
> Therefore I am loading the lib similar to `loadANARIDevice` would, check extensions and unload in the helper function.

Oh. I didn't appreciate that you wouldn't be able to use the same library/device loaded from `ANARILoadDevice`. I'm not a fan of checking the extensions from a different load than the device being used. Especially because `ANARILoadDevice` could potentially find a different library than `anari_cpp::loadLibrary`.

Maybe it's better we back out of the last change and just have the small fix for now.

marcelkrueger:
I understand your issue with this.
However, 
```cpp
ANARI_INTERFACE const char ** anariGetDeviceExtensions(ANARILibrary library, const char* deviceSubtype);
```
is a library level query. 
But as ANARILoadDevice ultimately sidesteps the library loading if it is a viskores device instantiating the device directly there is no library handle:
```cpp
anari_cpp::Device ANARILoadDevice(const std::string& libraryName)
{
  if (libraryName == "viskores")
  {
    VISKORES_LOG_F(viskores::cont::LogLevel::Info,
                   "Directly loading internal ANARI device named `viskores`");
    return reinterpret_cast<anari_cpp::Device>(
      new viskores_device::ViskoresDevice(AnariStatusFunc, nullptr));
  }
  else
  {
    VISKORES_LOG_S(viskores::cont::LogLevel::Info,
                   "Loading ANARI library named `" << libraryName << "`");
    anari_cpp::Library library =
      anari_cpp::loadLibrary(libraryName.c_str(), AnariStatusFunc, nullptr);
    if (library == nullptr)
    {
      VISKORES_LOG_S(viskores::cont::LogLevel::Info,
                     "Failed to load ANARI library named `" << libraryName << "`");
      return nullptr;
    }
    anari_cpp::Device device = anari_cpp::newDevice(library, "default");
    anari_cpp::unloadLibrary(library);
    if (device == nullptr)
    {
      VISKORES_LOG_S(viskores::cont::LogLevel::Info,
                     "Failed to load ANARI device from library named `" << libraryName << "`");
    }
    return device;
  }
}
```
Therefore,
```cpp
anariGetProperty(d, d, "extension", ANARI_STRING_LIST, ...)
```
should probably be the best bet if we only have the device.

At the moment I see two ways:
1) Leave it in the unit tests as is and add one to check the extensions on library level if you want to have regression tests against not reporting extensions
2) Refactor the unit tests in such a way that they also load or instantiate the library instead of a device if it's a viskores device.

But that can also be done in a different PR and I reverted it to the small fix for now.

kmorel:
> At the moment I see two ways:
> 
> 1. Leave it in the unit tests as is and add one to check the extensions on library level if you want to have regression tests against not reporting extensions

I think this is the better option going forward.

> But that can also be done in a different PR and I reverted it to the small fix for now.

Yes, this "small" change is ballooning into something bigger. Let's just merge in the small change and address the regression test later.

Oh, one more thing. Can you add a small file to `docs/changelog` documenting this fix? That's important for our release notes. You can mostly copy the explanation you put in comments in this PR.

marcelkrueger:
Added.

kmorel:
Looks good. Let's get the CI to pass and merge this in.

PR #348, opened 2026-06-04
Add second CI job using ANARI

kmorel:
Eventually, I would like to test both helide and viskores devices so we can compare our renderings to the reference implementation.

PR #366, opened 2026-06-30
Add Python bindings for Field, CoordinateSystem, DataSet, and DataSetBuilderUniform

ghweber:
## Summary

This PR exposes the core DataSet types to Python, building on the `UnknownArrayHandle` + NumPy interop from PR #308.

New bindings:

- **`viskores.Range` / `viskores.Bounds`** — mirror `viskores::Range` and `viskores::Bounds` with read/write `Min`/`Max` and `X`/`Y`/`Z` members plus `IsNonEmpty()`, `Length()`, and `Center()` accessors.
- **`viskores.cont.FieldAssociation`** — enum mirroring `Field::Association` (`Any`, `WholeDataSet`, `Points`, `Cells`, `Partitions`, `Global`)
- **`viskores.cont.Field(name, association, data)`** — wraps a `UnknownArrayHandle` with a name and association; `.GetData()` and `.asnumpy()` accessors
- **`viskores.cont.CoordinateSystem(name, data)`** — inherits `Field`; `.GetBounds()` returns a `viskores.Bounds`
- **`viskores.cont.DataSet`** — `AddField`, `AddPointField`, `AddCellField`, `HasField`, `GetField`, `AddCoordinateSystem`, `GetCoordinateSystem`, `GetNumberOf{Fields,Points,Cells,CoordinateSystems}`. Indexed accessors validate the index and raise instead of hitting undefined behavior (`VISKORES_ASSERT` is a no-op in release builds).
- **`viskores.cont.DataSetBuilderUniform.Create(dimensions)`** — accepts a 1-, 2-, or 3-element list of point counts, each >= 1. As in the C++ API, a dimension of 1 collapses that axis; dimensions < 1 are rejected because they would produce a dataset whose coordinate array disagrees with its cell set.

`DataSet.GetField()` returns by reference (`nb::rv_policy::reference_internal`); this is safe because `FieldCollection` stores fields in a `std::map`, so inserting new fields does not invalidate references to existing ones, and the policy keeps the DataSet alive while the Field is in use.

## Tests

- `PythonBindingsUnknownArrayHandleNumPy` — existing, still passes
- `PythonBindingsDataSet` — covers `FieldAssociation`, `Field`, uniform dataset creation (1D/2D/3D, collapsed axes, rejected bad dimensions), out-of-range index errors for `GetField`/`GetCoordinateSystem`, point/cell field round-trips, `HasField`, `GetField` by name and index, `Range`/`Bounds` construction and accessors, `CoordinateSystem` construction + `GetBounds()` from float32 and float64 NumPy arrays, and `asnumpy()` roundtrip for both float32 and float64 coordinate arrays

### DIFF
--- a/docs/changelog/python-bindings-dataset.md
+++ b/docs/changelog/python-bindings-dataset.md
@@ -11,11 +11,7 @@ build and inspect datasets from Python:
   `UnknownArrayHandle`; `.asnumpy()` is a convenience shortcut.
 
 * `viskores.cont.CoordinateSystem(name, data)` — constructs a coordinate
-  system from a name and a `UnknownArrayHandle`. Three-component float32 or
-  float64 arrays produced by `array_from_numpy` are copied to
-  `ArrayHandle<Vec3f_32>` / `ArrayHandle<Vec3f_64>` storage so that filters
-  can dispatch on them via `CoordinateSystem::GetData()`. See the design
-  question in `cont_bindings.cxx` for discussion of alternatives.
+  system from a name and a `UnknownArrayHandle`.
 
 * `viskores.cont.DataSet` — exposes `AddField`, `AddPointField`,
   `AddCellField`, `HasField`, `GetField`, `AddCoordinateSystem`,
--- a/docs/changelog/python-bindings-dataset.md
+++ b/docs/changelog/python-bindings-dataset.md
@@ -3,6 +3,10 @@
 The Viskores Python bindings now expose the core DataSet types needed to
 build and inspect datasets from Python:
 
+* `viskores.Range` and `viskores.Bounds` — mirror `viskores::Range` and
+  `viskores::Bounds` with read/write `Min`/`Max` and `X`/`Y`/`Z` members plus
+  `IsNonEmpty()`, `Length()`, and `Center()` accessors.
+
 * `viskores.cont.FieldAssociation` — enum with values `Any`, `WholeDataSet`,
   `Points`, `Cells`, `Partitions`, and `Global`.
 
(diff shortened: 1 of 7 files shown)


kmorel:
Before we start taking a look at this PR, let's work on adding Python testing to a couple of the CI jobs.

ghweber:
The two existing Python tests (`PythonBindingsUnknownArrayHandleNumPy` and `PythonBindingsDataSet`) are pure Python/NumPy API tests — they test data conversion and DataSet construction but don't dispatch any worklets or touch a device backend. They behave identically in a CPU-only build and a CUDA build.

For a CPU job, the plan is clear: add `python_bindings` to an existing job like `ubuntu2404_gcc14`, following the same pattern as ANARI in PR #348.

For the GPU job: the existing `ubuntu2404_gcc14_cuda130` has `no_testing` because CUDA device tests require real GPU hardware. Since the Python tests are backend-agnostic, they would pass on CPU runners even in a CUDA build — but that only verifies the bindings compile in a CUDA configuration rather than exercising any GPU code paths.

Should we defer the GPU CI job until there are Python tests that actually dispatch worklets, or is verifying compilation in a CUDA build sufficient for now?

kmorel:
> Should we defer the GPU CI job until there are Python tests that actually dispatch worklets, or is verifying compilation in a CUDA build sufficient for now?

I recommend doing both now. It will be less work all around to modify 2 in the same PR, and it will still be a pretty small PR.

kmorel:
I said I wouldn't really review until the CI tests are in place, but I saw this discussion that should be addressed.

> A Python-supplied coordinate array stored as `ArrayHandleRuntimeVec` will silently fail to dispatch in any filter that calls `CastAndCall` on coordinate data via `CoordinateSystem::GetData()`.

**False**. There are some special cases in `UnknownArrayHandle` that allow you to "downcast" to arrays that don't strictly match. This is the case for basic and runtime storage. For example, you can store an `ArrayHandleRuntimeVec<viskores::FloatDefault>` with 3 components in an `UnknownArrayHandle` and then retrieve it as an `ArrayHandle<viskores::Vec3f>`. The `CanConvert()`, `AsArrayHandle()`, and `CastAndCall()` functions all support this.

This is demonstrated in [an example in Section 4.8.4](https://viskores.readthedocs.io/en/latest/memory-layout.html#ex-groupwithruntimevec) in the user's guide. The text explanation is just above that.

So, the proper thing to do is nothing. But, I would add a test for this behavior by making this conversion while forcing copying off to verify that I am not lying.

PR #368, opened 2026-07-02
Add requirements file for Python bindings

kmorel:
This should make it easier to document what packages are needed for the Python bindings and easier for users and developers to install these dependencies.

### DIFF
--- a/CMake/ViskoresCheckLicense.cmake
+++ b/CMake/ViskoresCheckLicense.cmake
@@ -46,6 +46,7 @@ set(EXCEPTIONS
   DCO.txt
   README.txt
   docs/users-guide/requirements.txt
+  python_bindings/requirements.txt
   # Common directories with build files
   .venv
   build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,6 +347,64 @@ if (Viskores_ENABLE_TESTING)
 endif()
 
 #-----------------------------------------------------------------------------
+# Requirements for Python bindings
+if(Viskores_ENABLE_PYTHON_BINDINGS)
+  #-----------------------------------------------------------------------------
+  # Keep stable ABI support optional while the binding surface is small. This
+  # lets developers check abi3 compatibility without making it the default build.
+  option(VISKORES_PYTHON_STABLE_ABI "Build Python bindings with the CPython stable ABI" OFF)
+
+  set(Viskores_PYTHON_COMPONENTS Interpreter Development)
+  if (VISKORES_PYTHON_STABLE_ABI AND CMAKE_VERSION VERSION_GREATER_EQUAL 3.26)
+    list(APPEND Viskores_PYTHON_COMPONENTS Development.SABIModule)
+  endif()
+  find_package(Python 3.8 REQUIRED COMPONENTS ${Viskores_PYTHON_COMPONENTS})
+
+  #-----------------------------------------------------------------------------
+  # Load the nanobind library, which is used to create the Python binding libs.
+  file(STRINGS "${Viskores_SOURCE_DIR}/python_bindings/requirements.txt" _req_lines REGEX "^nanobind")
+  string(REGEX MATCH "([0-9]+\\.[0-9]+\\.[0-9]+)" Viskores_PYTHON_NANOBIND_MIN_VERSION "${_req_lines}")
+
+  execute_process(
+    COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
+    OUTPUT_VARIABLE _viskores_interop_python_nanobind_cmake_dir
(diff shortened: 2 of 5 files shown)


ghweber:
The PR looks good overall. One small maintenance risk: the `nanobind>=2.0.0` floor in `requirements.txt` is a second copy of whatever `Viskores_PYTHON_NANOBIND_MIN_VERSION` is set to in CMake. If someone bumps the CMake minimum later they'd need to remember to update `requirements.txt` too.

A few options for keeping them in sync:

**Option 1: Comment near the CMake variable**
Add a comment like `# Also update python_bindings/requirements.txt` next to `Viskores_PYTHON_NANOBIND_MIN_VERSION`. Low overhead, but relies on people noticing it.

**Option 2: CMake reads `requirements.txt`**
Make `requirements.txt` the authoritative source and have CMake parse it to set `Viskores_PYTHON_NANOBIND_MIN_VERSION`:
```cmake
file(STRINGS "${CMAKE_SOURCE_DIR}/python_bindings/requirements.txt" _req_lines
     REGEX "^nanobind")
string(REGEX MATCH "([0-9]+\\.[0-9]+\\.[0-9]+)" Viskores_PYTHON_NANOBIND_MIN_VERSION
       "${_req_lines}")
```
The file stays in the source tree, works standalone for `pip`, and CMake derives its minimum from it automatically. The tradeoff is added CMake complexity for a version that hasn't needed bumping often.

What's your preference?

kmorel:
I like option 2.

kmorel:
> I like option 2.

Oh, yeah. This is my PR.

I did a larger review of the issue and saw several things that could use consolidation. @ghweber, I overwrote your commit (we were both working at the same time) and consolidated several things including the Python loading and the version specification of numpy.

ghweber:
One potential edge case: the regex `([0-9]+\\.[0-9]+\\.[0-9]+)` extracts the first version-shaped string it finds, regardless of the operator. So `nanobind==2.0.0` or `nanobind~=2.0.0` would silently extract `2.0.0` and use it as a minimum — which may not be what was intended. Anchoring to `>=` and using `CMAKE_MATCH_1` makes the intent explicit and fails loudly if the operator is missing:

```cmake
string(REGEX MATCH ">=([0-9]+\\.[0-9]+\\.[0-9]+)" _ignored "${_req_lines}")
set(Viskores_PYTHON_NANOBIND_MIN_VERSION "${CMAKE_MATCH_1}")
if(NOT Viskores_PYTHON_NANOBIND_MIN_VERSION)
  message(FATAL_ERROR "Could not parse nanobind minimum version from requirements.txt. Expected 'nanobind>=X.Y.Z'.")
endif()
```

Same change applies to the numpy block. Up to you whether the extra robustness is worth the added lines.

vicentebolea:
I strongly recommend to go to the pyproject.toml route for this. This and using scikit-build-core to standardize pip installations

kmorel:
> One potential edge case

This is a rabbit hole that there is no point in following. There are a ton of little threads to pull: What if nanobind is not in requirements.txt? What if the patch version is left out of the version? If the minimum is not set, how does that verify the correctness?

Who cares? This is all just convenience to help set up a build. There is a comment in requirements.txt to specify a minimum version. That sounds easy enough to maintain.

kmorel:
> I strongly recommend to go to the pyproject.toml route for this. This and using scikit-build-core to standardize pip installations

I'm too lazy to do that. There is no president in Viskores to attempt to set up a user's build environment (other than another convenience requirements.txt in the users-guide directory). What you describe sounds like something to put in a package manager.

vicentebolea:
> > I strongly recommend to go to the pyproject.toml route for this. This and using scikit-build-core to standardize pip installations
> 
> I'm too lazy to do that. There is no president in Viskores to attempt to set up a user's build environment (other than another convenience requirements.txt in the users-guide directory). What you describe sounds like something to put in a package manager.

The concept is that it would allow to build this package by doing `pip install .`. I will also allow making viskores wheels and sdist. This come right on time with an effort for VTK wheels now trying to depends as much as possible to its core dependencies through PIP.

kmorel:
> > > I strongly recommend to go to the pyproject.toml route for this. This and using scikit-build-core to standardize pip installations
> > 
> > 
> > I'm too lazy to do that. There is no president in Viskores to attempt to set up a user's build environment (other than another convenience requirements.txt in the users-guide directory). What you describe sounds like something to put in a package manager.
> 
> The concept is that it would allow to build this package by doing `pip install .`. I will also allow making viskores wheels and sdist. This come right on time with an effort for VTK wheels now trying to depends as much as possible to its core dependencies through PIP.

OK, I've replaced this with #387. It adds a `pyproject.toml` file and synchronizes it with a `requirements.txt` file.

Issue #331, opened 2026-05-27
Add a Rotational Extrusion filter


nicolemarsaglia:
Ascent would love to be able to revolve data around an axis, and maybe even an arbitrary line, if we're being greedy.

kmorel:
Can you be more specific about what you need?

The [`PointTransform`] filter can perform arbitrary affine transformations. It comes with convenience functions to define rotations around an arbitrary axis of rotation. That axis of rotation has to be through the origin, but you can also set an arbitrary transformation matrix. You can use the functions in `viskores/Transform3D.h` to translate to the origin, rotate, and translate back to effectively rotate around an arbitrary point.

If you want it to just look like the object is rotating without changing the points, you could rotate the camera around it.

[`PointTransform`]: https://viskores.readthedocs.io/en/latest/provided-filters.html#point-transform

nicolemarsaglia:
We don't just want to move the data. Specifically we want to take a 2D rz mesh and revolve it to create a 3D representation.  

And turns all our use cases will be revolving around an axis.

kmorel:
> We don't just want to move the data. Specifically we want to take a 2D rz mesh and revolve it to create a 3D representation.

Oh. You want a rotational extrusion filter. I updated the issue title to reflect that.

A linear extrusion filter would be nice, too. It would be a simple addition once the  mechanics of the rotational extrusion is designed.

nicolemarsaglia:
Yes, thank you!

meghanto:
<img width="1914" height="1294" alt="Image" src="https://github.com/user-attachments/assets/9751936e-1933-4e23-a8c4-f2bb89902889" />

I implemented something like this as a viskores filter. is this what you want @nicolemarsaglia ? then I can add it as a PR

For context, the semi-transparent volume is the result after rotation is performed, the solid cell/cells are the initial profile given as input

nicolemarsaglia:
@meghanto this is exactly what we want! Thanks for working on getting this added!

PR #369, opened 2026-07-07
Add rotational and linear extrusion

meghanto:
This branch adds `LinearExtrusion` and `RotationalExtrusion` filters to `viskores::filter::geometry_refinement`. (as requested in https://github.com/Viskores/viskores/issues/331 ) 

Both filters sweep a triangulated 2D profile into wedge cells. The shared `AbstractExtrusion` implementation handles input validation, optional triangulation, compact or explicit output topology, and point/cell field mapping. Its implementation lives in `AbstractExtrusion.hxx`. Separate worklets generate the linear and rotational coordinates. 

For now, this assumes triangle cells as extrusion profiles, but I think this can be extended to other cell types with minimal changes. 

`LinearExtrusion` translates the profile along a direction for a specific distance. `RotationalExtrusion` revolves the profile around an axis with configurable center, sweep angle, and behavior for closed-sweep (connecting ending plane with the starting plane). Full `2*pi` rotational sweeps default to closed unless explicitly overridden.

This also adds unit coverage for both extrusion filters and a changelog note.

Full test suites passed. 

🤖 Assisted by OpenAI Codex

### DIFF
--- a/viskores/filter/geometry_refinement/CMakeLists.txt
+++ b/viskores/filter/geometry_refinement/CMakeLists.txt
@@ -17,6 +17,9 @@
 ##============================================================================
 set(geometry_refinement_headers
   ConvertToPointCloud.h
+  ExtrusionAbstract.h
+  ExtrusionLinear.h
+  ExtrusionRotational.h
   Shrink.h
   SplitSharpEdges.h
   Tetrahedralize.h
@@ -27,6 +30,8 @@ set(geometry_refinement_headers
 
 set(geometry_refinement_sources
   ConvertToPointCloud.cxx
+  ExtrusionLinear.cxx
+  ExtrusionRotational.cxx
   Shrink.cxx
   SplitSharpEdges.cxx
   Tetrahedralize.cxx
@@ -35,9 +40,14 @@ set(geometry_refinement_sources
   VertexClustering.cxx
   )
 
+set(geometry_refinement_template_sources
+  ExtrusionAbstract.hxx
+  )
+
 viskores_library(
   NAME viskores_filter_geometry_refinement
   HEADERS ${geometry_refinement_headers}
+  TEMPLATE_SOURCES ${geometry_refinement_template_sources}
   DEVICE_SOURCES ${geometry_refinement_sources}
   USE_VISKORES_JOB_POOL
 )
--- a/viskores/filter/geometry_refinement/ExtrusionLinear.h
+++ b/viskores/filter/geometry_refinement/ExtrusionLinear.h
@@ -1,0 +1,55 @@
+//============================================================================
+//  The contents of this file are covered by the Viskores license. See
+//  LICENSE.txt for details.
+//
+//  By contributing to this file, all contributors agree to the Developer
+//  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+//============================================================================
+
+#ifndef viskores_filter_geometry_refinement_ExtrusionLinear_h
+#define viskores_filter_geometry_refinement_ExtrusionLinear_h
+
+#include <viskores/filter/geometry_refinement/ExtrusionAbstract.h>
+#include <viskores/filter/geometry_refinement/viskores_filter_geometry_refinement_export.h>
+
+namespace viskores
+{
+namespace filter
+{
+namespace geometry_refinement
+{
+
+/// @brief Extrude a triangulated profile along a direction into wedge cells.
+///
+/// `ExtrusionLinear` creates a sequence of translated copies of the input points
+/// along a direction and connects adjacent triangle copies with wedge cells.
+class VISKORES_FILTER_GEOMETRY_REFINEMENT_EXPORT ExtrusionLinear
+  : public viskores::filter::geometry_refinement::ExtrusionAbstract
+{
+public:
+  VISKORES_CONT ExtrusionLinear() { this->SetNumberOfPlanes(2); }
+
+  /// @brief Set the extrusion direction.
+  ///
+  /// The direction vector is normalized before use. Use `SetDistance` to set
+  /// the spacing between adjacent planes.
+  VISKORES_CONT void SetDirection(viskores::Vec3f direction) { this->Direction = direction; }
+  /// @copydoc SetDirection
+  VISKORES_CONT viskores::Vec3f GetDirection() const { return this->Direction; }
+
+  /// @brief Set the extrusion distance along the normalized direction.
+  VISKORES_CONT void SetDistance(viskores::FloatDefault distance) { this->Distance = distance; }
+  /// @copydoc SetDistance
+  VISKORES_CONT viskores::FloatDefault GetDistance() const { return this->Distance; }
+
+private:
+  VISKORES_CONT viskores::cont::DataSet DoExecute(const viskores::cont::DataSet& input) override;
+
+  viskores::Vec3f Direction = { 0, 0, 1 };
+  viskores::FloatDefault Distance = 1;
+};
+} // namespace geometry_refinement
+} // namespace filter
+} // namespace viskores
+
+#endif // viskores_filter_geometry_refinement_ExtrusionLinear_h
--- a/viskores/filter/geometry_refinement/testing/UnitTestExtrusionLinear.cxx
+++ b/viskores/filter/geometry_refinement/testing/UnitTestExtrusionLinear.cxx
@@ -1,0 +1,290 @@
+//============================================================================
+//  The contents of this file are covered by the Viskores license. See
+//  LICENSE.txt for details.
+//
+//  By contributing to this file, all contributors agree to the Developer
+//  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
+//============================================================================
+
+#include <viskores/CellShape.h>
+#include <viskores/cont/CellSetExtrude.h>
+#include <viskores/cont/CellSetSingleType.h>
+#include <viskores/cont/DataSetBuilderExplicit.h>
+#include <viskores/cont/ErrorFilterExecution.h>
+#include <viskores/cont/testing/Testing.h>
+#include <viskores/filter/geometry_refinement/ExtrusionLinear.h>
+
+#include <array>
+#include <functional>
+#include <vector>
+
+namespace
+{
+
+using ExplicitWedgeCellSet = viskores::cont::CellSetSingleType<>;
+
+viskores::cont::DataSet MakeTriangleDataSet()
+{
+  const std::vector<viskores::Vec3f> coordinates{ { 0.0f, 0.0f, 0.0f }, { 1.0f, 0.0f, 0.0f },
+                                                  { 0.0f, 1.0f, 0.0f }, { 2.0f, 0.0f, 0.0f },
+                                                  { 3.0f, 0.0f, 0.0f }, { 2.0f, 1.0f, 0.0f } };
+  const std::vector<viskores::UInt8> shapes{ viskores::CELL_SHAPE_TRIANGLE,
+                                             viskores::CELL_SHAPE_TRIANGLE };
+  const std::vector<viskores::IdComponent> numIndices{ 3, 3 };
+  const std::vector<viskores::Id> connectivity{ 0, 1, 2, 3, 5, 4 };
+
+  viskores::cont::DataSetBuilderExplicit builder;
+  viskores::cont::DataSet dataSet =
+    builder.Create(coordinates, shapes, numIndices, connectivity, "coords");
+
+  dataSet.AddPointField("pointvar",
+                        std::vector<viskores::Float32>{ 10.0f, 20.0f, 30.0f, 40.0f, 50.0f, 60.0f });
+  dataSet.AddCellField("cellvar", std::vector<viskores::Float32>{ 70.0f, 80.0f });
+  return dataSet;
+}
+
+viskores::cont::DataSet MakeQuadDataSet()
+{
+  const std::vector<viskores::Vec3f> coordinates{
+    { 0.0f, 0.0f, 0.0f }, { 1.0f, 0.0f, 0.0f }, { 1.0f, 1.0f, 0.0f }, { 0.0f, 1.0f, 0.0f }
+  };
+  const std::vector<viskores::UInt8> shapes{ viskores::CELL_SHAPE_QUAD };
+  const std::vector<viskores::IdComponent> numIndices{ 4 };
+  const std::vector<viskores::Id> connectivity{ 0, 1, 2, 3 };
+
+  viskores::cont::DataSetBuilderExplicit builder;
+  return builder.Create(coordinates, shapes, numIndices, connectivity, "coords");
+}
+
+bool ThrowsFilterExecution(const std::function<void()>& function)
+{
+  try
+  {
+    function();
+  }
+  catch (const viskores::cont::ErrorFilterExecution&)
+  {
+    return true;
+  }
+  return false;
+}
+
+void AssertCellPointIds(const viskores::cont::DataSet& dataSet,
+                        viskores::Id cellId,
+                        const std::array<viskores::Id, 6>& expected)
+{
+  std::array<viskores::Id, 6> actual;
+  dataSet.GetCellSet().GetCellPointIds(cellId, actual.data());
+  for (viskores::IdComponent index = 0; index < 6; ++index)
+  {
+    VISKORES_TEST_ASSERT(actual[static_cast<std::size_t>(index)] ==
+                           expected[static_cast<std::size_t>(index)],
+                         "Wrong wedge point id.");
+  }
+}
+
+viskores::Float64 SignedWedgeVolume(const viskores::cont::DataSet& dataSet, viskores::Id cellId)
+{
+  std::array<viskores::Id, 6> pointIds;
+  dataSet.GetCellSet().GetCellPointIds(cellId, pointIds.data());
+
+  viskores::cont::ArrayHandle<viskores::Vec3f> coordinates;
+  dataSet.GetCoordinateSystem().GetData().AsArrayHandle(coordinates);
+  auto coordinatePortal = coordinates.ReadPortal();
+
+  const viskores::Vec3f point0 = coordinatePortal.Get(pointIds[0]);
+  const viskores::Vec3f point1 = coordinatePortal.Get(pointIds[1]);
+  const viskores::Vec3f point2 = coordinatePortal.Get(pointIds[2]);
+  const viskores::Vec3f point3 = coordinatePortal.Get(pointIds[3]);
+  const viskores::Vec3f point4 = coordinatePortal.Get(pointIds[4]);
+  const viskores::Vec3f point5 = coordinatePortal.Get(pointIds[5]);
+
+  viskores::Float64 volume =
+    viskores::Dot(viskores::Cross(point1 - point0, point2 - point0), point3 - point0);
+  volume += viskores::Dot(viskores::Cross(point4 - point1, point5 - point1), point3 - point1);
+  volume += viskores::Dot(viskores::Cross(point5 - point1, point2 - point1), point3 - point1);
+  return volume;
+}
+
+void AssertPositiveWedgeVolumes(const viskores::cont::DataSet& dataSet)
+{
+  for (viskores::Id cellId = 0; cellId < dataSet.GetNumberOfCells(); ++cellId)
+  {
+    VISKORES_TEST_ASSERT(SignedWedgeVolume(dataSet, cellId) > 0.0,
+                         "Output wedge should have positive orientation.");
+  }
+}
+
+void TestDefaultOutput()
+{
+  std::cout << "Testing default linear extrusion output" << std::endl;
+
+  viskores::filter::geometry_refinement::ExtrusionLinear filter;
+  viskores::cont::DataSet output = filter.Execute(MakeTriangleDataSet());
+
+  VISKORES_TEST_ASSERT(output.GetNumberOfPoints() == 12, "Wrong default output point count.");
+  VISKORES_TEST_ASSERT(output.GetNumberOfCells() == 2, "Wrong default output cell count.");
+  VISKORES_TEST_ASSERT(output.GetCellSet().IsType<ExplicitWedgeCellSet>(),
+                       "Default output cell set is not explicit single-type topology.");
+
+  const auto cellSet = output.GetCellSet().AsCellSet<ExplicitWedgeCellSet>();
+  VISKORES_TEST_ASSERT(cellSet.GetCellShapeAsId() == viskores::CELL_SHAPE_WEDGE,
+                       "Default output cells should be wedges.");
+  VISKORES_TEST_ASSERT(cellSet.GetNumberOfPointsInCell(0) == 6,
+                       "Output wedges should have 6 points.");
+  AssertCellPointIds(output, 0, { 0, 1, 2, 6, 7, 8 });
+  AssertCellPointIds(output, 1, { 3, 4, 5, 9, 10, 11 });
+  AssertPositiveWedgeVolumes(output);
+}
+
+void TestMultiplePlanes()
+{
+  std::cout << "Testing multiple linear extrusion planes" << std::endl;
+
+  viskores::filter::geometry_refinement::ExtrusionLinear filter;
+  filter.SetNumberOfPlanes(4);
+  filter.SetDirection({ 0.0f, 0.0f, 2.0f });
+  filter.SetDistance(3.0f);
+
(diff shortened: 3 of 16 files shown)


viskores/filter/geometry_refinement/testing/UnitTestLinearExtrusion.cxx
```diff
@@ -0,0 +1,273 @@
...
+
+//============================================================================
+//  Copyright (c) Kitware, Inc.
+//  All rights reserved.
+//  See LICENSE.txt for details.
+//
+//  This software is distributed WITHOUT ANY WARRANTY; without even
+//  the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+//  PURPOSE.  See the above copyright notice for more information.
+//============================================================================
```
kmorel: We don't need to add this copyright statement to new files. New files just need the statement at the top. It is better to not add this statement to new files. (This applies to all new files.)

viskores/filter/geometry_refinement/testing/UnitTestLinearExtrusion.cxx
```diff
@@ -0,0 +1,273 @@
...
+using ExplicitWedgeCellSet = viskores::cont::CellSetSingleType<>;
+
+viskores::cont::DataSet MakeTriangleDataSet()
+{
+  const std::vector<viskores::Vec3f> coordinates{ { 0.0f, 0.0f, 0.0f },
+                                                  { 1.0f, 0.0f, 0.0f },
+                                                  { 0.0f, 1.0f, 0.0f } };
+  const std::vector<viskores::UInt8> shapes{ viskores::CELL_SHAPE_TRIANGLE };
+  const std::vector<viskores::IdComponent> numIndices{ 3 };
+  const std::vector<viskores::Id> connectivity{ 0, 1, 2 };
```
kmorel: Let's make this data set have two triangles with normals pointing in opposite directions to make sure that the filter sets the windings correctly.

```suggestion
  const std::vector<viskores::Vec3f> coordinates{ { 0.0f, 0.0f, 0.0f },
                                                  { 1.0f, 0.0f, 0.0f },
                                                  { 0.0f, 1.0f, 0.0f },
                                                  { 1.0f, 1.0f, 0.0f } };
  const std::vector<viskores::UInt8> shapes{ viskores::CELL_SHAPE_TRIANGLE };
  const std::vector<viskores::IdComponent> numIndices{ 3 };
  const std::vector<viskores::Id> connectivity{ 0, 1, 2, 1, 2, 3 };
```

viskores/filter/geometry_refinement/testing/UnitTestLinearExtrusion.cxx
```diff
@@ -0,0 +1,273 @@
...
+
+  viskores::cont::DataSetBuilderExplicit builder;
+  viskores::cont::DataSet dataSet =
+    builder.Create(coordinates, shapes, numIndices, connectivity, "coords");
+
+  dataSet.AddPointField("pointvar", std::vector<viskores::Float32>{ 10.0f, 20.0f, 30.0f });
+  dataSet.AddCellField("cellvar", std::vector<viskores::Float32>{ 40.0f });
```
kmorel: To match the previous change.

```suggestion
  dataSet.AddPointField("pointvar", std::vector<viskores::Float32>{ 10.0f, 20.0f, 30.0f, 40.0f });
  dataSet.AddCellField("cellvar", std::vector<viskores::Float32>{ 50.0f });
```

viskores/filter/geometry_refinement/testing/UnitTestLinearExtrusion.cxx
```diff
@@ -0,0 +1,273 @@
...
+
+void TestDefaultOutput()
+{
+  std::cout << "Testing default linear extrusion output" << std::endl;
+
+  viskores::filter::geometry_refinement::LinearExtrusion filter;
+  viskores::cont::DataSet output = filter.Execute(MakeTriangleDataSet());
+
+  VISKORES_TEST_ASSERT(output.GetNumberOfPoints() == 6, "Wrong default output point count.");
+  VISKORES_TEST_ASSERT(output.GetNumberOfCells() == 1, "Wrong default output cell count.");
```
kmorel: To match the changes in dataset creation.

```suggestion
  VISKORES_TEST_ASSERT(output.GetNumberOfPoints() == 8, "Wrong default output point count.");
  VISKORES_TEST_ASSERT(output.GetNumberOfCells() == 2, "Wrong default output cell count.");
```

viskores/filter/geometry_refinement/testing/UnitTestExtrusionLinear.cxx
```diff
@@ -0,0 +1,273 @@
...
+  VISKORES_TEST_ASSERT(output.GetNumberOfPoints() == 6, "Wrong default output point count.");
+  VISKORES_TEST_ASSERT(output.GetNumberOfCells() == 1, "Wrong default output cell count.");
+  VISKORES_TEST_ASSERT(output.GetCellSet().IsType<ExplicitWedgeCellSet>(),
+                       "Default output cell set is not explicit single-type topology.");
+
+  const auto cellSet = output.GetCellSet().AsCellSet<ExplicitWedgeCellSet>();
+  VISKORES_TEST_ASSERT(cellSet.GetCellShapeAsId() == viskores::CELL_SHAPE_WEDGE,
+                       "Default output cells should be wedges.");
+  VISKORES_TEST_ASSERT(cellSet.GetNumberOfPointsInCell(0) == 6,
+                       "Output wedges should have 6 points.");
```
kmorel, line 134: Should check the winding of both generated cells to make sure the normal of the first triangle points toward the second triangle.

viskores/filter/geometry_refinement/testing/UnitTestLinearExtrusion.cxx
```diff
@@ -0,0 +1,273 @@
...
+
+  const auto cellSet = output.GetCellSet().AsCellSet<viskores::cont::CellSetExtrude>();
+  VISKORES_TEST_ASSERT(!cellSet.GetIsPeriodic(), "Linear compact output should not be periodic.");
+  VISKORES_TEST_ASSERT(cellSet.GetCellShape(0) == viskores::CELL_SHAPE_WEDGE,
+                       "Compact output cells should be wedges.");
+}
+
+void TestClosedCompactOutput()
+{
+  std::cout << "Testing closed compact linear extrusion output" << std::endl;
```
kmorel: Does a closed sweep make sense in a linear extrusion?

viskores/filter/geometry_refinement/testing/UnitTestRotationalExtrusion.cxx
```diff
@@ -0,0 +1,368 @@
...
+{
+
+using ExplicitWedgeCellSet = viskores::cont::CellSetSingleType<>;
+
+viskores::cont::DataSet MakeTriangleDataSet(bool includeAxisPoint = false)
+{
+  std::vector<viskores::Vec3f> coordinates;
+  if (includeAxisPoint)
+  {
+    coordinates = { { 0.0f, 0.0f, 0.0f }, { 2.0f, 0.0f, 0.0f }, { 1.0f, 0.0f, 1.0f } };
```
kmorel: I won't repeat all the comments from the linear extrusion test, but like that test it would be good to have two triangles with opposite windings and test that the resulting wedges have the correct point order.

viskores/filter/geometry_refinement/worklet/Extrusion.h
```diff
@@ -0,0 +1,98 @@
...
+struct ExtrusionOutputToInputCell
+{
+  viskores::Id NumberOfCellsPerPlane = 0;
+
+  VISKORES_EXEC_CONT
+  viskores::Id operator()(viskores::Id outputCellId) const
+  {
+    return outputCellId % this->NumberOfCellsPerPlane;
+  }
+};
```
kmorel: This is not a big deal, but I notice that the implementation of these two structs are identical. They can be merged into one.


```suggestion
struct ExtrusionOutputToInputPlane
{
  viskores::Id NumberOfIdsPerPlane = 0;

  VISKORES_EXEC_CONT
  viskores::Id operator()(viskores::Id outputId) const
  {
    return outputId % this->NumberOfIdsPerPlane;
  }
};
```

viskores/filter/geometry_refinement/CMakeLists.txt
```diff
@@ -16,7 +16,10 @@
 ##  PURPOSE.  See the above copyright notice for more information.
 ##============================================================================
 set(geometry_refinement_headers
+  AbstractExtrusion.h
   ConvertToPointCloud.h
+  LinearExtrusion.h
+  RotationalExtrusion.h
```
kmorel: Viskores naming convention prefers that these classes (and associated files) would be named `ExtrusionLinear` and `ExtrusionRotational` (and `ExtrusionAbstract`). Putting descriptors from general to specific makes classes easy to find when, say, you are looking at an alphabetical list of files.

viskores/filter/geometry_refinement/worklet/RotationalExtrusion.h
```diff
@@ -0,0 +1,99 @@
...
+    const viskores::Vec<ComponentType, 3> rotated =
+      center + parallel + (perpendicular * viskores::Cos(theta)) +
+      (viskores::Cross(axis, perpendicular) * viskores::Sin(theta));
+
+    output = rotated;
+  }
+
+private:
+  viskores::Vec3f Axis;
+  viskores::Vec3f Center;
```
kmorel: Since these get cast to `ComponentType` in the operator, it might be best to make these `viskores::Vec3f_64`. That way you won't run into precision problems with compiling with default 32-bit floats and you encounter a 64-bit float field.

viskores/filter/geometry_refinement/AbstractExtrusion.h
```diff
@@ -0,0 +1,106 @@
...
+namespace geometry_refinement
+{
+namespace internal
+{
+
+VISKORES_CONT inline void ThrowExtrusionExecution(const std::string& filterName,
+                                                  const std::string& message)
+{
+  throw viskores::cont::ErrorFilterExecution(filterName + ": " + message);
+}
```
kmorel: This is an unnecessary function that an overly enthusiastic LLM created. We can get rid of this and just throw an exception directly.

viskores/filter/geometry_refinement/AbstractExtrusion.h
```diff
@@ -0,0 +1,106 @@
...
+
+VISKORES_CONT inline viskores::Float64 MagnitudeSquared64(const viskores::Vec3f& value)
+{
+  const viskores::Float64 x = static_cast<viskores::Float64>(value[0]);
+  const viskores::Float64 y = static_cast<viskores::Float64>(value[1]);
+  const viskores::Float64 z = static_cast<viskores::Float64>(value[2]);
+  return (x * x) + (y * y) + (z * z);
+}
```
kmorel: This is probably an unnecessary function. In the whole 2 places its used, you can just cast the value to float-64 and use the Viskores-provided magnitude squared.

```cpp
viskores::Float64 mag2 = viskores::MagnitudeSquared(viskores::Vec3f_64{ value });
```

viskores/filter/geometry_refinement/AbstractExtrusion.hxx
```diff
@@ -0,0 +1,357 @@
...
+    {
+      ThrowExtrusionExecution(filterName, "input cells must be triangles.");
+    }
+    connectivity =
+      cellSet.GetConnectivityArray(viskores::TopologyElementTagCell{},
+                                   viskores::TopologyElementTagPoint{});
+  }
+  else
+  {
+    ThrowExtrusionExecution(filterName, "input cell set must be explicit triangle topology.");
```
kmorel: One day we may want to implement the extrusion of a CellSetStructured<2>, but that is a PR for another day.

viskores/filter/geometry_refinement/worklet/Extrusion.h
```diff
@@ -0,0 +1,98 @@
...
+    const viskores::Id cellId = flatConnectivityId / 6;
+    const viskores::Id localPointId = flatConnectivityId % 6;
+    const viskores::Id triangleId = cellId % this->NumberOfCellsPerPlane;
+    const viskores::Id plane0 = cellId / this->NumberOfCellsPerPlane;
+    const viskores::Id plane1 = (plane0 < (this->NumberOfPlanes - 1)) ? (plane0 + 1) : 0;
+    const viskores::Id basePointId = baseConnectivity.Get((triangleId * 3) + (localPointId % 3));
+
+    output = (localPointId < 3)
+      ? ((plane0 * this->NumberOfPointsPerPlane) + basePointId)
+      : ((plane1 * this->NumberOfPointsPerPlane) + nextNode.Get(basePointId));
```
kmorel: A problem with this function is that it might get the "winding" wrong for the wedge. If you look closely at the [point ordering for the wedge cell], you will see that the order of points for the triangle face defined by 0, 1, 2 should be ordered such that "front" side of the face (determined by counterclockwise traversal) should point to the other triangular face. This function gets it right if the input triangle's normal points in the direction of extrusion but wrong if pointing the other way. Getting the winding wrong could cause downstream computation like volumes and gradients to be computed incorrectly.

To get this right, each triangle normal would have to be compared to the direction of extrusion. If they face different directions, the first and second triangles should swap planes.

[point ordering for the wedge cell]: https://viskores.readthedocs.io/en/latest/dataset.html#tab-explicitcellsetshapes

viskores/filter/geometry_refinement/AbstractExtrusion.hxx
```diff
@@ -0,0 +1,357 @@
...
+  {
+    internal::ThrowExtrusionExecution(filterName, "open sweeps require at least 2 planes.");
+  }
+
+  const viskores::Id maxInt32 =
+    static_cast<viskores::Id>(std::numeric_limits<viskores::Int32>::max());
+  if (this->NumberOfPlanes > maxInt32)
+  {
+    internal::ThrowExtrusionExecution(filterName,
+                                      "number of planes exceeds CellSetExtrude Int32 limits.");
```
kmorel: There are several checks and error throws to ensure `NumberOfPlanes` does not exceed a 32-bit integer. Why is that? Granted, that many planes could create problematically large data, but otherwise I don't see anything about this implementation that would be limited to 32-bit integers. Why all these checks?

viskores/filter/geometry_refinement/AbstractExtrusion.hxx
```diff
@@ -0,0 +1,357 @@
...
+  const viskores::Id numberOfPoints = workingInput.GetCellSet().GetNumberOfPoints();
+  if (inputCoordinates.GetData().GetNumberOfValues() != numberOfPoints)
+  {
+    internal::ThrowExtrusionExecution(filterName,
+                                      "coordinate count does not match cell set point count.");
+  }
+  if (numberOfPoints > maxInt32)
+  {
+    internal::ThrowExtrusionExecution(filterName,
+                                      "number of input points exceeds CellSetExtrude Int32 limits.");
```
kmorel: Same comment about number of points. This seems like an unnecessary limitation.

viskores/filter/geometry_refinement/RotationalExtrusion.cxx
```diff
@@ -0,0 +1,166 @@
...
+    const viskores::Float64 dot = (axisX * x) + (axisY * y) + (axisZ * z);
+    const viskores::Float64 perpendicularX = x - (dot * axisX);
+    const viskores::Float64 perpendicularY = y - (dot * axisY);
+    const viskores::Float64 perpendicularZ = z - (dot * axisZ);
+    const viskores::Float64 perpendicular2 = (perpendicularX * perpendicularX) +
+      (perpendicularY * perpendicularY) + (perpendicularZ * perpendicularZ);
+
+    return perpendicular2 <= this->Tolerance2;
+  }
+};
```
kmorel: If you make the `Axis` and `Center` fields 64-bit, you can simplify this implementation quite a bit.


```suggestion
struct IsPointOnAxis
{
  viskores::Vec3f_64 Axis;
  viskores::Vec3f_64 Center;
  viskores::Float64 Tolerance2;

  template <typename PointType>
  VISKORES_EXEC_CONT bool operator()(const PointType& pointValue) const
  {
    const viskores::Vec3f_64 relativeP = viskores::Vec3f_64{ pointValue } - this->Center;

    const viskores::Float64 dot = viskores::Dot(this->Axis, relativeP);
    const viskores::Vec3f_64 perpendicular = relativeP - (dot * this->Axis);
    const viskores::Float64 perpendicularMag2 = viskores::MagnitudeSquared(perpendicular);

    return perpendicularMag2 <= this->Tolerance2;
  }
};
```

viskores/filter/geometry_refinement/RotationalExtrusion.cxx
```diff
@@ -0,0 +1,166 @@
...
+    internal::ThrowExtrusionExecution("RotationalExtrusion",
+                                      "axis point tolerance must be nonnegative.");
+  }
+  if (this->GetCloseSweep() && !NearlyFullSweep(this->SweepAngle))
+  {
+    internal::ThrowExtrusionExecution("RotationalExtrusion",
+                                      "closed sweeps require a full 2*pi sweep angle.");
+  }
+
+  const viskores::Float64 inverseAxisMagnitude = 1.0 / viskores::Sqrt(axisMagnitude2);
```
kmorel:
```suggestion
  const viskores::Float64 inverseAxisMagnitude = viskores::Rsqrt(axisMagnitude2);
```

kmorel (changes requested):
Thanks for contributing this new feature, @meghanto.

I have several comments with suggested changes. The biggest issue I see is that the extrusion has the potential for creating inverted cells. Maybe we can initially ignore that issue, but in that case it should be documented that care should be taken to ensure that normals face correctly.

Speaking of documentation, please add some documentation to the provided filters chapter in the user's guide. This is in `viskores/docs/users-guide/provided-filters.rst`. Codex is pretty good about creating a section with a simple prompt.

viskores/filter/geometry_refinement/AbstractExtrusion.h
meghanto: Fixed!

viskores/filter/geometry_refinement/AbstractExtrusion.h
meghanto: Fixed!

viskores/filter/geometry_refinement/worklet/RotationalExtrusion.h
meghanto: Fixed!

viskores/filter/geometry_refinement/CMakeLists.txt
meghanto: Fixed!

viskores/filter/geometry_refinement/worklet/Extrusion.h
meghanto: Fixed!

viskores/filter/geometry_refinement/testing/UnitTestLinearExtrusion.cxx
meghanto: https://github.com/Viskores/viskores/pull/369/changes/a3de3b5340df29f83ef21dbba09b6ab6dbd5ee73 fixes this by moving closed sweep a rotational-only feature. While the method ExecuteTriangleExtrusion in ExtrusionAbstract.hxx requires a closesweep boolean, the ExtrusionLinear class passes a false when it calls it.

viskores/filter/geometry_refinement/RotationalExtrusion.cxx
meghanto: Fixed!

viskores/filter/geometry_refinement/RotationalExtrusion.cxx
meghanto: Fixed!

meghanto:
### Re: inverted cells

Addressed by orienting the per-triangle connectivity before constructing the wedges.  Comparing the triangle normal against the local sweep direction, we swap the second and third triangle point ids in the oriented connectivity used for output construction if the initial orientation would lead to inverted cells.

Instead of swapping planes, we swap the initial connectivity because it keeps local wedge points 0,1,2 on the current plane and 3,4,5 on the next plane, which matches the CellSetExtrude representation and keeps the point correspondence between adjacent planes as identity (fast computation). Also keeps closed sweeps simple because the final connection wraps to plane 0 with the same point ids.

### Re: Int32 checks

The general extrusion path no longer has Int32 limits. Input connectivity and explicit wedge output now use viskores::Id.

The remaining Int32 checks are only inside the compact-output branch because viskores::cont::CellSetExtrude itself stores its topology as Int32: connectivity, nextNode, numberOfPointsPerPlane, and numberOfPlanes are all Int32 in CellSetExtrude. So those checks are guarding the conversion required to construct compact CellSetExtrude output. If compact output is not requested, the explicit output path does not apply those Int32 limits.

viskores/filter/geometry_refinement/worklet/Extrusion.h
```diff
@@ -0,0 +1,126 @@
...
+                                const CoordinatePortalType& baseCoordinates,
+                                viskores::Id3& output) const
+  {
+    using CoordinateType = typename CoordinatePortalType::ValueType;
+    using ComponentType = typename CoordinateType::ComponentType;
+
+    const viskores::Id connectivityOffset = triangleId * 3;
+    const viskores::Id pointId0 = baseConnectivity.Get(connectivityOffset + 0);
+    const viskores::Id pointId1 = baseConnectivity.Get(connectivityOffset + 1);
+    const viskores::Id pointId2 = baseConnectivity.Get(connectivityOffset + 2);
```
kmorel: This implementation is fine, and you don't have to change anything, but I want to note that there is a trick to make the worklet a bit simpler. Instead of having a `triangleId` array and a separate connectivity array, you can just read in the connectivity array as a simple `FieldIn`. Then, on the control/invoke side, the connectivity array can be wrapped in an [`ArrayHandleGroupVec`] to group every 3 entries into a `viskores::Id3`. Then, the worklet just counts by triangles and gets the `viskores::Id3` containing all three point ids for the triangle.

[`ArrayHandleGroupVec`]: https://viskores.readthedocs.io/en/latest/fancy-array-handles.html#grouped-vector-arrays

viskores/filter/geometry_refinement/ExtrusionLinear.h
```diff
@@ -0,0 +1,52 @@
...
+///
+/// `ExtrusionLinear` creates a sequence of translated copies of the input points
+/// along a direction and connects adjacent triangle copies with wedge cells.
+class VISKORES_FILTER_GEOMETRY_REFINEMENT_EXPORT ExtrusionLinear
+  : public viskores::filter::geometry_refinement::ExtrusionAbstract
+{
+public:
+  VISKORES_CONT ExtrusionLinear() { this->SetNumberOfPlanes(2); }
+
+  /// @brief Set the extrusion direction.
```
kmorel, line 32: The documentation string should make clear that the direction vector gets normalized, and users must use `SetDistance` to specify the spacing between planes.

kmorel (changes requested):
Thanks for making those changes. This PR is looking very good and close to being ready to merge. As long as the CI comes back clean, there are only a few further things to do.

1. Make the minor change to the doxygen comment pointed to in this review.
2. ~The formatting test failed.~ Never mind. I see you have already fixed that.
3. The commits should be merged together. This can be done with either [rebase](https://github.com/Viskores/viskores/blob/main/CONTRIBUTING.md#rebase-to-squash-commits) or [reset](https://github.com/Viskores/viskores/blob/main/CONTRIBUTING.md#reset-to-squash-commits).
4. Fix any further issues with the CI.

Once that is done, we can merge.

viskores/filter/geometry_refinement/ExtrusionLinear.h
meghanto: Thanks, didn't consider this, fixed!

viskores/filter/geometry_refinement/worklet/Extrusion.h
meghanto: I made the change because it makes sense and goes better with the rest of the codebase!

meghanto:
Should be ready for final CI check now. Last time only an unrelated kokkos test failed, hopefully every one of them passes now.

kmorel (approved):
Looks good. This will be merged in once the tests pass.

kmorel:
> Should be ready for final CI check now. Last time only an unrelated kokkos test failed, hopefully every one of them passes now.

Ping me if it looks like an unrelated test fails. Sometimes tests need to be restarted.

PR #371, opened 2026-07-09
Make viskores_interop_python a header-only interface library

ghweber:
## Summary

Converts `viskores_interop_python` from a static `viskores_library` to a header-only `INTERFACE` library, per [discussion on #367](https://github.com/Viskores/viskores/pull/367#issuecomment-4920171742).

- `ArrayHandleToNumPy()` and `NumPyToArrayHandle()` move from `.cxx` files into their headers as `inline` functions.
- `viskores_interop_python` becomes a plain `add_library(... INTERFACE)` target with no compiled sources, exported via `viskores_install_targets()`.
- The now-unnecessary `viskores_interop_python_export.h` and `VISKORES_INTEROP_PYTHON_EXPORT` macro are removed.

### Why

`viskores_interop_python`'s entry points take and return `nb::object`/`nb::handle`, so every conceivable consumer is itself a nanobind extension module compiling with nanobind headers. nanobind types are only interoperable across extension modules that share an identical ABI tag (nanobind version, stable-ABI setting, compiler), and there is no shared nanobind runtime to link against — every consumer compiles nanobind into its own module. A compiled (static/shared/object) `viskores_interop_python` would therefore tie every consumer to the exact nanobind build used for Viskores itself — the concern originally raised on #308 when the library was made static.

Header-only removes that coupling entirely: each consumer compiles the conversion code against its own nanobind runtime. It also makes the library usable from outside the Viskores build for the first time — an `OBJECT` library (the shape used on the `python-ci-testing` branch to fix a Linux `BUILD_SHARED_LIBS=ON` link failure) exports as an `INTERFACE` target with no object files unless the raw `.o` files are installed, so external `find_package(Viskores)` consumers got undefined references. Now `viskores::interop_python` is a proper `INTERFACE IMPORTED` target with `INTERFACE_LINK_LIBRARIES viskores::cont`.

The compiled surface being given up is small: all the template machinery (`RuntimeVecArrayViewOwner`, `PyBufferOwner`, dtype dispatch tables) was already header-only; only the two non-template entry points (~100 lines total) move back into headers. Per-class Python bindings (#366 and beyond) live in the extension module's own sources and are unaffected.

## Test plan

Verified locally (macOS/AppleClang, Python 3.14, nanobind 2.13, numpy 2.5) with `Viskores_ENABLE_PYTHON_BINDINGS=ON`:

- [x] Configure + build with `BUILD_SHARED_LIBS=OFF` — builds cleanly, `PythonBindingsUnknownArrayHandleNumPy.py` passes.
- [x] Configure + build with `BUILD_SHARED_LIBS=ON` — builds cleanly, same test passes. (This is the configuration that originally motivated the `OBJECT` library workaround on #367; header-only sidesteps the underlying linker-ordering issue entirely since there's no archive to order.)
- [x] `cmake --install` to a scratch prefix:
  - Both interop headers installed; no stale `viskores_interop_python_export.h`.
  - `viskores_config.mk` still contains no `-lviskores_interop_python` flag.
  - `ViskoresTargets.cmake` exports `viskores::interop_python` as `INTERFACE IMPORTED` with `INTERFACE_LINK_LIBRARIES "viskores::cont"`.
- [x] Configure with `Viskores_ENABLE_PYTHON_BINDINGS=OFF` — no `interop_python` target is created; module stays disabled as before.

This PR should merge before #367 (Python bindings CI); that branch will rebase on top and drop its `OBJECT`-library commit, keeping only the CI configuration and the separate `MODULE_LIBRARY` `lib_args` exclusion (needed for `viskores_python_bindings` itself, unrelated to this change).

https://claude.ai/code/session_01TvvcG2m6nEzcn81tjUMQaZ


viskores/interop/python/CMakeLists.txt
```diff
@@ -6,55 +6,21 @@
...
-  message(FATAL_ERROR
-    "nanobind ${Viskores_PYTHON_NANOBIND_MIN_VERSION} or newer is required to "
-    "build viskores_interop_python. Install with: pip install 'nanobind>="
-    "${Viskores_PYTHON_NANOBIND_MIN_VERSION}'.")
-endif()
-
-find_package(nanobind ${Viskores_PYTHON_NANOBIND_MIN_VERSION} CONFIG REQUIRED
-  PATHS "${_viskores_interop_python_nanobind_cmake_dir}"
-  NO_DEFAULT_PATH
-)
```
kmorel, line 29: We should still find the `nanobind` library and make it a public link library to `viskores_interop_python`. Anything that compiled the previous .cxx needed them, so now anything that includes the header files needs them. (It probably should have been a public link in the first place as nanobind headers were always included from these headers.)

kmorel (changes requested):

viskores/interop/python/CMakeLists.txt
ghweber: Good catch that the header requirement is currently implicit — you're right that it shouldn't be left to luck. But when I went to implement "find nanobind and make it a public link library," it turned out not to be directly possible, and the reason is worth laying out because it also explains why header-only is the right approach here.

**`find_package(nanobind)` doesn't create a linkable target.** I read through the full [`nanobind-config.cmake`](https://github.com/wjakob/nanobind/blob/master/cmake/nanobind-config.cmake) — every `add_library` in it is inside a function body, so at `find_package` time nothing is defined except the `NB_DIR`/`NB_SUFFIX` variables and the helper functions. There is no `nanobind::nanobind` or headers target (the only importable target the config defines is `tsl::robin_map`). So there's nothing to `target_link_libraries(viskores_interop_python PUBLIC …)` against.

**The one nanobind library target is created as a side effect of `nanobind_add_module()`**, not `find_package`. It's `EXCLUDE_FROM_ALL`, and its name varies with build flags (`nanobind-static`, `nanobind-static-abi3`, `-ft`, `-ps`, …). Linking it into `viskores_interop_python` — which we export in `ViskoresTargets.cmake` — would either fail at export time (it's not in any install set) or hard-code this machine's `NB_DIR` path and Python include dirs into the installed target, pointing external consumers at the specific nanobind/Python environment Viskores happened to be built with.

This is actually the same coupling that motivated going header-only. nanobind's [FAQ on type visibility](https://nanobind.readthedocs.io/en/latest/faq.html#type-visibility) spells out that nanobind types only interoperate across extensions when they "use the same nanobind ABI version… the same compiler… ABI-compatible versions of the C++ library… the stable ABI interface consistently… debug/release mode consistently." Any prebuilt form of this library ties every consumer to Viskores's exact nanobind build; header-only lets each consumer compile the conversion code against its own nanobind, so the constraint dissolves. It also lines up with nanobind's own guidance to consolidate nanobind includes and type casters into a shared header used across binding translation units — which is essentially what `viskores_interop_python` now is. (nanobind's [documented helper-library pattern](https://nanobind.readthedocs.io/en/latest/api_cmake.html) is only for project code that *doesn't* include nanobind headers, which is why it links cleanly there and doesn't map onto our case.)

**One scope clarification, so "usable by external code" isn't over-read.** What makes this safe to hand to an arbitrary external nanobind consumer is that the library's interchange format is NumPy arrays, not nanobind-bound types: `ArrayHandleToNumPy` returns a `numpy.ndarray` and `NumPyToArrayHandle` takes one. A `numpy.ndarray` is a plain CPython object with no nanobind-ABI dependency, and the C++ `UnknownArrayHandle` side is shared through the normal `viskores_cont` library — so nothing crosses a nanobind runtime boundary, and a consumer built against a different nanobind is fine on this path. (In fact this is where header-only is strictly safer than a prebuilt lib would be: a prebuilt `viskores_interop_python` would ship conversion code compiled against *our* nanobind and force external consumers onto our exact ABI even for the numpy path.) What this does *not* promise is cross-module sharing of Viskores's own bound types — handing a `viskores.cont.DataSet` object from `_viskores` to an external module and unwrapping it back to C++ — which requires matching nanobind ABI plus a shared `NB_DOMAIN`. That constraint is inherent to the `_viskores` bound types regardless of how this library is implemented, and nobody depends on it yet; I just want to be clear the header-only change isn't quietly claiming to solve it.

That said, your underlying point stands: the target should carry its header requirement explicitly rather than working only because `nanobind_add_module` happens to supply the includes to the one current consumer. I think the right way to express that — without leaking build-machine paths into the install — is a build-tree-only interface requirement:

```cmake
find_package(Python 3.8 REQUIRED COMPONENTS Interpreter Development)
find_package(nanobind CONFIG REQUIRED …)   # same discovery python_bindings already does

add_library(viskores_interop_python INTERFACE)
target_link_libraries(viskores_interop_python INTERFACE viskores_cont)
target_include_directories(viskores_interop_python INTERFACE
  $<BUILD_INTERFACE:${NB_DIR}/include>
  $<BUILD_INTERFACE:${Python_INCLUDE_DIRS}>)
viskores_install_targets(TARGETS viskores_interop_python)
```

In-tree consumers then get nanobind/Python headers from the target directly instead of by accident, while external consumers still resolve nanobind in their own environment (via their own `find_package(nanobind)` / `nanobind_add_module`), which is what they have to do regardless. The `$<BUILD_INTERFACE:>` guard is what keeps the absolute paths out of the exported `ViskoresTargets.cmake`.

Does that approach work for you? Happy to push it if so.

kmorel: The LLM is being pretty chatty, huh?

Fair point, and the point is well taken. This proposed solution seems hacky to me. Let's go back to the solution we started with (no linking to nanobind/python at all). As previously pointed out, anything using this library is necessarily using nanobind on its own. Let that code find its own nanobind.

ghweber: I didn't trust it completely and wasn't sure whether you would trust it. So I explicitly asked it to not just make statements but also include the reasoning and sources.

(I am trying to be more questioning and push back on it making statements.)

ghweber: Sounds good — keeping it as-is with no nanobind/Python linkage (the current branch). Could you clear the review when you get a chance so this can merge ahead of #367?

kmorel (approved):

PR #375, opened 2026-07-12
Make wireframe overlay visibile

kmorel:
There was an issue when combining overlay wireframes with pseudocolor renderings where the pseudocolor rendering would obscure the wireframe. Viskores now detects when a wireframe is rendered as an overlay and ignores the depth buffer to ensure visibility.

### DIFF
--- a/docs/changelog/wireframe-visibility.md
+++ b/docs/changelog/wireframe-visibility.md
@@ -1,0 +1,6 @@
+## Wireframe overlay now visible through pseudocolor
+
+There was an issue when combining overlay wireframes with pseudocolor renderings
+where the pseudocolor rendering would obscure the wireframe. Viskores now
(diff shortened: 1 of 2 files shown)


kmorel:
@nicolemarsaglia, please review this wireframer fix from the patch you sent me.

Note that I have not included the changes to `MapperWireframer`. These changes changes seem to just check to see if an unstructured grid has only polygons and skip the external faces extraction. I removed it because (1) this change is independent from the occlusion fix and should be in a separate PR, (2) the check uses an inefficient loop of cells on the host, and (3) if we really want this feature it should be integrated as an option in the `ExternalFaces` filter.

kmorel:
@vicentebolea We would like this change in the 1.2 release to support Ascent.

nicolemarsaglia (approved):
Thanks for getting this in!

PR #376, opened 2026-07-12
Fix rendering issue with viewport changes

kmorel:
Viskores had an issue with setting the viewport properly when not in 3D camera mode. The camera now properly resizes the viewport when using 2D camera projections.

PR #377, opened 2026-07-12
Fix rendering issue with viewport changes - backport v12

kmorel:
Viskores had an issue with setting the viewport properly when not in 3D camera mode. The camera now properly resizes the viewport when using 2D camera projections.


kmorel:
@nicolemarsaglia This is a duplicate PR of #376 to push it to the 1.2 release instead of main. It just needs a check when you review the other one.

vicentebolea (approved):

Issue #374, opened 2026-07-10
HistSampling computes per-partition histograms instead of a global histogram


meghanto:
## Motivation

I have been revisiting some of my past research work to complete contributions that never made it fully upstream. `HistSampling` was originally moved from Ascent into VTK-m in [VTK-m !3083](https://gitlab.kitware.com/vtk/vtk-m/-/merge_requests/3083), but Ascent still carries a separate implementation. I last worked on this in 2023 (!) during the vtk-m hackathon at ANL ([Alpine-DAV/ascent#1200](https://github.com/Alpine-DAV/ascent/pull/1200)).

While investigating whether Ascent could use the Viskores filter directly, I found that partitioned execution does not use the global histogram required by the sampling algorithm.

## Problem

`HistSampling` overrides only `DoExecute(const DataSet&)`. The default `PartitionedDataSet` execution therefore runs the complete algorithm independently on each partition.

Each partition computes its own field range, histogram, sample target, and acceptance probabilities. This makes the sampling distribution depend on partition boundaries. A value that is common globally can be treated as rare within one partition. The same problem applies across MPI ranks.

The algorithm instead requires one histogram over the complete population. Viskores already supports the necessary partitioned and MPI reduction in `density_estimate::Histogram`.

## Proposed change

Add `HistSampling::DoExecutePartitions` so that it:

1. Computes one histogram over all partitions and MPI ranks.
2. Derives one global acceptance-probability array and sample target.
3. Applies that distribution to each partition.
4. Returns the sampled point-cloud partitions with their fields preserved.

Tests could verify that rarity and acceptance probabilities are based on the global population rather than individual partitions.

This issue is limited to correcting partitioned point-field sampling. Cell-associated sampling and conversion to cell-centered points were also part of the earlier PR, but can be handled in a follow-up change.

Would this be an appropriate direction for a future PR?

kmorel:
Yes, I agree that this would be a better implementation of `HistSampling`. This sounds like a straightforward change.

PR #384, opened 2026-07-16
Use a global histogram for partitioned HistSampling

meghanto:
This addresses #374.

### Summary

- Compute one histogram across all partitions and MPI ranks before applying `HistSampling`.
- Reuse the base partition scheduler when applying the shared sampling probabilities.
- Preserve empty `PartitionedDataSet` behavior, including ranks with no local partitions.
- Document the global partitioned behavior and add local, threaded, empty-input, and MPI coverage.

### Implementation

`HistSampling::DoExecutePartitions` now builds a distributed histogram, derives one acceptance-probability array from its global bin counts, and delegates partition processing to the base filter machinery.

A collective partition-count check returns early only when the input is empty across every rank. A rank with no local partitions therefore still participates in the histogram collectives when another rank has data. This adds one scalar all-reduce per partitioned execution in MPI builds, separate from the existing global range and histogram reductions.

Single-`DataSet` behavior is unchanged.

### Testing

- `UnitTestHistSampling`, including partition-independent rarity, threaded execution, and empty input
- `UnitTestHistSamplingMPI` with three ranks, including uneven partition distribution, a rank with no local partitions, and globally empty input
- 100 repeated serial-device threaded runs and 20 repeated three-rank MPI runs on a compute node

### DIFF
--- a/viskores/filter/resampling/HistSampling.cxx
+++ b/viskores/filter/resampling/HistSampling.cxx
@@ -70,18 +70,23 @@ viskores::cont::ArrayHandle<viskores::FloatDefault> CalculatPdf(
 
 viskores::cont::DataSet HistSampling::DoExecute(const viskores::cont::DataSet& input)
 {
-  //computing histogram based on input
-  viskores::filter::density_estimate::Histogram histogram;
-  histogram.SetNumberOfBins(this->NumberOfBins);
-  histogram.SetActiveField(this->GetActiveFieldName());
-  auto histogramOutput = histogram.Execute(input);
-  viskores::cont::ArrayHandle<viskores::Id> binCountArray;
-  viskores::cont::ArrayCopyShallowIfPossible(
-    histogramOutput.GetField(histogram.GetOutputFieldName()).GetData(), binCountArray);
-  viskores::Id totalPoints = input.GetNumberOfPoints();
-  //computing pdf
-  viskores::cont::ArrayHandle<viskores::FloatDefault> probArray;
-  probArray = CalculatPdf(totalPoints, this->SampleFraction, binCountArray);
+  if (!this->InExecutePartitions)
+  {
+    //computing histogram based on input
+    viskores::filter::density_estimate::Histogram histogram;
+    histogram.SetNumberOfBins(this->NumberOfBins);
+    histogram.SetActiveField(this->GetActiveFieldName(), this->GetActiveFieldAssociation());
+    auto histogramOutput = histogram.Execute(input);
+    viskores::cont::ArrayHandle<viskores::Id> binCountArray;
+    viskores::cont::ArrayCopyShallowIfPossible(
+      histogramOutput.GetField(histogram.GetOutputFieldName()).GetData(), binCountArray);
+    viskores::Id totalPoints = input.GetNumberOfPoints();
+    //computing pdf
+    this->Probabilities = CalculatPdf(totalPoints, this->SampleFraction, binCountArray);
+    this->HistogramMinimum = histogram.GetComputedRange().Min;
+    this->HistogramBinDelta = histogram.GetBinDelta();
+  }
+
   // use the acceptance probabilities and random array to create 0-1 array
   // generating random array between 0 to 1
   viskores::cont::ArrayHandle<viskores::Int8> outputArray;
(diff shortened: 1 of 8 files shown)


viskores/filter/density_estimate/Histogram.cxx
```diff
@@ -247,14 +267,51 @@ VISKORES_CONT void Histogram::PreExecute(const viskores::cont::PartitionedDataSe
   }
   else
   {
-    auto handle = viskores::cont::FieldRangeGlobalCompute(
-      input, this->GetActiveFieldName(), this->GetActiveFieldAssociation());
```
kmorel: As near as I can tell, all the changes to the Histogram filter just add threaded computation to compute the local histogram. In this case, two changes should be made.

1. These improvements should be implemented in `FieldRangeGlobalCompute.cxx`. If they are useful for this filter, they will be useful elsewhere.
2. This change should be made in a separate PR. The changes to `HistSample` will work with or without the changes to how field ranges are computed.

viskores/filter/density_estimate/testing/UnitTestPartitionedDataSetHistogramFilter.cxx
```diff
@@ -113,6 +112,8 @@ static void TestPartitionedDataSetHistogram()
 
   viskores::filter::density_estimate::Histogram histogram;
   histogram.SetActiveField("double");
+  histogram.SetThreadsPerCPU(2);
+  histogram.SetRunMultiThreadedFilter(runMultiThreaded);
```
kmorel: It's probably OK, but it is not typical for outside code to set the threads for a filter. It's mostly for the concrete filter to communicate with its superclass based on whether the filter is thread safe.

That said, this was implemented but not very well studied. There are plenty of optimizations to be had.

@dpugmire Feel free to comment more on using the threaded multiblock feature.

meghanto:
Checking to see if this requires some more work, please let me know!

kmorel (changes requested):
Sorry, @meghanto. I reviewed this code a while ago, but my review never got submitted. I think sometimes I click the GItHub `Submit review` button that does not actually submit the review.

Anyway, here is my review. I have a requested code restructuring to allow the parallel global range to be available everywhere.

viskores/filter/density_estimate/Histogram.cxx
meghanto: Agreed. I split the threaded Histogram work out of this PR and moved the reusable local-partition implementation into FieldRangeGlobalCompute in #432. This PR now contains only the global HistSampling behavior and its coverage.

viskores/filter/density_estimate/testing/UnitTestPartitionedDataSetHistogramFilter.cxx
meghanto: I removed these settings from this PR with the rest of the threaded Histogram work. In #432 they are used only as deterministic test controls to compare the sequential and two-worker paths; production Histogram execution obtains its device-aware worker count from the base filter scheduler.

meghanto:
Does this PR need any more changes before merge?

viskores/filter/resampling/HistSampling.cxx
```diff
@@ -70,18 +70,23 @@ viskores::cont::ArrayHandle<viskores::FloatDefault> CalculatPdf(
...
-  viskores::Id totalPoints = input.GetNumberOfPoints();
-  //computing pdf
-  viskores::cont::ArrayHandle<viskores::FloatDefault> probArray;
-  probArray = CalculatPdf(totalPoints, this->SampleFraction, binCountArray);
+  if (!this->InExecutePartitions)
+  {
+    //computing histogram based on input
+    viskores::filter::density_estimate::Histogram histogram;
+    histogram.SetNumberOfBins(this->NumberOfBins);
+    histogram.SetActiveField(this->GetActiveFieldName());
```
kmorel: This is a very minor thing, but I noticed in `PreExecute` you added a correction to pass the active association in addition to the field name. That same correction should probably be here in the single partition case.

```suggestion
    histogram.SetActiveField(this->GetActiveFieldName(), this->GetActiveFieldAssociation());
```

kmorel (changes requested):
Looks good. I just had one very minor suggested change.

viskores/filter/resampling/HistSampling.cxx
meghanto: Fixed!

kmorel (approved):
Great. Now we just need to CI to pass. Send me a note if there is a problem with that.

meghanto:
One CI fails on UnitTestProbe with NumericalError

kmorel:
That's probably an unrelated intermittent failure. I tried restarting.

PR #329, opened 2026-05-21
Add ANARIDevice management object

kmorel:
The anari interop library now contains a class named `ANARIDevice` that manages an ANARI device. In addition to making sure resources get created and destroyed correctly, it can find the built in Viskores device without having to rely on the user's environment.

PR #385, opened 2026-07-16
Fix ANARI loaded-device lifetime

jeffamstutz:
Keep dynamically loaded libraries alive in a move-only owner until the owned device is released. The previous loader unloaded the module before returning its device, leaving the live handle backed by an unloaded implementation.

Update ANARI test callers and add an optional real-backend lifetime regression.

### DIFF
--- a/docs/changelog/anari-loaded-device-lifetime.md
+++ b/docs/changelog/anari-loaded-device-lifetime.md
@@ -1,0 +1,23 @@
+## Keep dynamically loaded ANARI devices valid
+
+`viskores::interop::anari::ANARILoadDevice` now returns a move-only
+`ANARILoadedDevice` owner instead of a raw `anari_cpp::Device`. The owner keeps
+the dynamic ANARI library loaded for the lifetime of its device, then releases
+the device before unloading the library.
+
+Keep the returned owner alive while using its raw device handle:
+
+```cpp
+auto loadedDevice = viskores::interop::anari::ANARILoadDevice("helide");
+if (loadedDevice)
+{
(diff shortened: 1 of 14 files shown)


kmorel:
So, my first implementation of `ANARILoadDevice` was pretty similar to this. Instead of a function, it was an object that initialized a device. (One main difference was it used a `std::shared_ptr` so it could be copied around.)

My implementation had the `ANARILoadedDevice` passed around in place of the device and had a cast operator to be used in place of a device. Thus, anything keeping a reference to a device would ensure that the device was kept alive. However, I abandoned this because I was essentially making a wrapper library around ANARI to manage all its references. But the ANARI SDK is what it is, and I don't want to second-guess and redefine everything.

I get the point of making resource management classes, but if ANARI is not doing it, why should we? I see how this is helpful, but could get in the way in places like this:

```cpp
#ifdef USING_VISKORES
  auto device = viskores::interop::anari::ANARILoadDevice(...);
#else
  auto library = anari_cpp::loadLibrary(...);
  auto device = anari_cpp::newDevice(...);
  anari_cpp::unloadLibrary(library);
#endif
```

jeffamstutz:
Yeah, you're right -- going in that direction of wrapping ANARI is not great. I'll reform this into verifying that the tests have the correct library lifetimes.

kmorel:
> Yeah, you're right -- going in that direction of wrapping ANARI is not great. I'll reform this into verifying that the tests have the correct library lifetimes.

If you're interested, here is my abandoned PR: #329.

I ended up change every single class in the interop before realizing I was adding yet another interface layer that no one was asking for.

PR #387, opened 2026-07-17
Add pip installation support for Python bindings

kmorel:
## Summary

- add a `pyproject.toml` backed by scikit-build-core so the Viskores Python bindings can be installed with `pip install .`
- consolidate Python dependencies in the repository-level requirements file and derive CMake minimum versions from the package metadata
- build pip wheels with TBB support on supported Linux and Windows x86-64 platforms
- add CI coverage that keeps shared dependency constraints synchronized

## Why

The Python bindings previously required a manual CMake setup and maintained dependency constraints in multiple locations. Providing standard Python package metadata makes installation simpler, while the consistency check prevents the pip and developer requirements from drifting apart.

## Impact

Users can install the Python bindings and dependencies directly from the repository with `pip install .`. Existing CMake-based development builds remain supported.

## Validation

- `python3 Utilities/Scripts/check-python-dependencies.py`
- `git diff --check origin/main...HEAD`


kmorel:
@vicentebolea Here is an implementation using `pyproject.toml`. I tried to make this a little better than useless by adding TBB support where available, but even that parallelism is not available everywhere.

More generally, I don't know how we can get `python install .` to compile for available devices (particularly GPUs), and that limits the value of installing it this way.

vicentebolea:
> @vicentebolea Here is an implementation using `pyproject.toml`. I tried to make this a little better than useless by adding TBB support where available, but even that parallelism is not available everywhere.
> 
> More generally, I don't know how we can get `python install .` to compile for available devices (particularly GPUs), and that limits the value of installing it this way.

Sounds good, as for the GPUs, traditionally it couldnt be done with pip, so we needed conda for this. In the past year or two the cuda-toolkit is avaiable in PIP. In principle we should just list it as a dependency and that should be enough.

As for the TBB, they miss some packages for some arch/platforms.

I will followup with the PR review shortly

kmorel:
@vicentebolea ping

kmorel:
@vicentebolea Ping

vicentebolea:
Sorry @kmorel it took me a while to review this, the key concern that I see here is that we still have two paths to generate the python bindings, one is with `pip install` (thru the pyproject.toml) and another is the traditional `cmake . | cmake --install...`. These results in workarounds such as the script that you added to conciliate possible different versions of deps between requirement.txt and pyproject.toml

I strongly recommend to only support the pyproject.toml path (standard python way) and let go the `requirement.txt` and the workarounds to conciliate it happen. Note that we can still manually build the bindings using cmake only given that we have the required deps, we can setup CI jobs to do this if needed, we could also explore vendoring nanobind in viskores.

Good catch to also enable ABI3 compatible Python in nanobind.

I also recommend to eventually bring the pipeline of wheel python wheels and sdist build, test and upload from the ADIOS2 project to here to properly test this and possibly start pushing wheels to pypi.

kmorel:
@vicentebolea 

I really, really want to have a `requirements.txt` in addition to the `pip install .` for multiple reasons.

1. As discussed earlier, `pip install .` does not cover all use cases, and we are not sure how to make it cover all use cases. When we want a build that supports GPUs, we still need to go back to compiling with CMake.
2. Even if it did cover all use cases, developers still need to be able to compile the bindings via CMake and their IDE for development/debugging purposes. It doesn't matter whether `cmake --install` works or not. That's not the reason I'm building the binding or making the dependencies.
3. Even if developers didn't need to compile the bindings with CMake, the `requirements.txt` contains dependencies used outside of what is used for `pip install .` (e.g., user's guide, pyexpander, and CI replication).

The `pip install .` thing is feature creep on this PR. It solves a completely different problem than what I was originally trying to address, and I really just want to have a simple way to set up my Python binding for building and be done with it.

CMakeLists.txt
```diff
@@ -346,6 +346,64 @@ if (Viskores_ENABLE_TESTING)
...
 
+#-----------------------------------------------------------------------------
+# Requirements for Python bindings
+if(Viskores_ENABLE_PYTHON_BINDINGS)
+  #-----------------------------------------------------------------------------
+  # Keep stable ABI support optional while the binding surface is small. This
+  # lets developers check abi3 compatibility without making it the default build.
+  option(VISKORES_PYTHON_STABLE_ABI "Build Python bindings with the CPython stable ABI" OFF)
+
+  set(Viskores_PYTHON_COMPONENTS Interpreter Development)
```
vicentebolea: Numpy can be added here

vicentebolea:
I see, fair point I keep pushing for the pyproject.toml to support `pip install .` but that was not the main goal of this PR. having the requirement.txt helps users setup the required deps when going the direct cmake route. 

> When we want a build that supports GPUs, we still need to go back to compiling with CMake.

This used to be the case, however, these days, there is the cuda-sdk package in Pypi but still your point stands with other more exotic gpu/accelerators backends.

2 and 3 are very fair point.

requirements.txt
```diff
@@ -5,73 +5,62 @@
 ##  By contributing to this file, all contributors agree to the Developer
 ##  Certificate of Origin Version 1.1 (DCO 1.1) as stated in DCO.txt.
 ##============================================================================
+
```
vicentebolea, line 8: I recommend keep these in two diff txt files.

You could do requirements.txt (for core) and requirements-doc.txt (for the sphinx) and `pip install -r requirements.txt` will actually read the both of the file (-r flag).

vicentebolea:
I left two comments, everything else looks good

requirements.txt
kmorel: I have split `requirements.txt` into multiple files. I've placed the different `requirements.txt` files in subdirectories next to the place where they go (in some cases restoring where they came from).

In the `requirements.txt` file at the repo root, I use `-r` flags to load in all the sub-requirements with one call.

kmorel:
@vicentebolea I updated based on your suggestions. Please re-review.

vicentebolea (approved):

PR #391, opened 2026-07-21
Remove old copyright

kmorel:
When Viskores was rebranded from VTK-m and joined the HPSF as part of the Linux Foundation, the license statement required at the top of each file changed. Rather than provide the specifics of the copyright in every file, the new statement just points to LICENSE.txt and declares that every contribution implicitly agrees with the DCO.

The old statements were originally left in the source files as they are not incompatible with the new statement (and contributors are allowed to add their own copyright statements in addition to our license statement.) However, because the outdated statement was still so prevalent, it kept getting added to new files (particularly files created by LLMs). To prevent this, we are removing those old statements.


kmorel:
@vicentebolea, @berkgeveci Are you OK with these changes? This should not be a big deal because the copyright and license have not really changed, but I want to make sure Kitware is OK with this.

kmorel:
@vicentebolea @berkgeveci As a reminder, I would like some feedback on whether this copyright removal is appropriate.

vicentebolea:
> @vicentebolea @berkgeveci As a reminder, I would like some feedback on whether this copyright removal is appropriate.

I have consulted this internally with our legal/compliance team and I am now waiting for their advise on it.

vicentebolea (approved):
I have consulted this internally and there should be no problem with this change since it does not change the license of the source files since this change still links every source file with the centralized License.

kmorel:
Thanks. I'll make sure the CI is happy and merge in.

PR #397, opened 2026-07-23
Add testing chapter to user's guide

kmorel:


kmorel:
@vicentebolea Here is another short user's guide chapter for your review. Built chapter can be viewed at https://viskores--397.org.readthedocs.build/en/397/testing.html.

vicentebolea (approved):
Looks and renders well

PR #406, opened 2026-08-06
Use plain CMake option()

dg0yt:
cmake_minimum_required is 3.15.

Noticed due to missing help text in CMakeCache.txt.


vicentebolea (approved):
Thanks

vicentebolea:
For some reason the CI does not start, it does not give the option to approve the workflow either..

vicentebolea:
@dg0yt can you amend and re-push?

dg0yt:
There are queued jobs... Maybe we need to give more time.
If nothing else helps I could also close and reopen the PR.

kmorel:
Weird. I just clicked the approve workflows to run button, and it appeared to work. But now everything seems stuck. Maybe we just need to wait a little bit... ?

PR #412, opened 2026-08-07
Ensure the depth passed to ANARI is the Euclidean distance to the camera

DeVasconcelos:
It's really exciting that there is a viskores device for ANARI now!

While playing around with it, I've noticed, however, that the depth buffer channel of the viskores device does not seem to align with the definition in the [ANARI specification](https://registry.khronos.org/ANARI/specs/1.1/ANARI-1.1.html#:~:text=channel%2Edepth) which states that `channel.depth` should be the "euclidean distance to the camera (not to the image plane)".

To resolve this (and to ensure that, from a user side, different ANARI devices do not have to be treated differently), I have added a buffer called `DistancesToCamera` to the `Canvas` class which is filled with the minimum distances passed to the `WriteToCanvas` methods in the `SurfaceConverter` worklet. This buffer is then passed to ANARI in the `map` method of the `Frame` class instead of the original `DepthBuffer`.

Please let me know what you think of this change.

### DIFF
--- a/viskores/rendering/Canvas.h
+++ b/viskores/rendering/Canvas.h
@@ -99,7 +99,7 @@ public:
 
   /// @brief Change the size of the image.
   VISKORES_CONT
-  void ResizeBuffers(viskores::Id width, viskores::Id height);
+  virtual void ResizeBuffers(viskores::Id width, viskores::Id height);
 
   /// @brief Specify the background color.
   VISKORES_CONT
--- a/docs/changelog/pass-euclidean-distance-to-camera-to-anari.md
+++ b/docs/changelog/pass-euclidean-distance-to-camera-to-anari.md
@@ -1,0 +1,3 @@
+## Use the Euclidean distances to the camera as depth buffer for ANARI
+
+To adhere to the ANARI specification, Viskores now sets the `channel.depth` parameter of the ANARI camera to the Euclidean distances to the camera.
--- a/viskores/rendering/CanvasRayTracer.cxx
+++ b/viskores/rendering/CanvasRayTracer.cxx
@@ -23,6 +23,18 @@ namespace rendering
 namespace internal
 {
 
+struct ClearDistancesToCamera : public viskores::worklet::WorkletMapField
+{
+  using ControlSignature = void(FieldOut);
+  using ExecutionSignature = void(_1);
+
+  VISKORES_CONT
+  ClearDistancesToCamera() {}
+
+  VISKORES_EXEC
+  void operator()(viskores::Float32& distance) const { distance = viskores::Infinity32(); }
+}; // struct ClearDistancesToCamera
+
 class SurfaceConverter : public viskores::worklet::WorkletMapField
 {
   viskores::Matrix<viskores::Float32, 4, 4> ViewProjMat;
@@ -36,9 +48,15 @@ public:
   {
   }
 
-  using ControlSignature =
-    void(FieldIn, WholeArrayInOut, FieldIn, FieldIn, FieldIn, WholeArrayInOut, WholeArrayInOut);
-  using ExecutionSignature = void(_1, _2, _3, _4, _5, _6, _7, WorkIndex);
+  using ControlSignature = void(FieldIn,
+                                WholeArrayInOut,
+                                FieldIn,
+                                FieldIn,
+                                FieldIn,
+                                WholeArrayInOut,
+                                WholeArrayInOut,
+                                WholeArrayInOut);
+  using ExecutionSignature = void(_1, _2, _3, _4, _5, _6, _7, _8, WorkIndex);
   template <typename Precision,
             typename ColorPortalType,
             typename DepthBufferPortalType,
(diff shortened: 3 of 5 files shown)


kmorel:
@DeVasconcelos, thanks for trying out the Viskores ANARI device and fixing issues with it. This is one of I'm sure several places that our device does not match up perfectly with the spec. We could use several people beating on it to get it the right shape.

Your implementation looks correct, but I want to poke at it a bit. This implementation adds another buffer to the ray casting that needs to be updated each intersection. But what if instead the ray casting was left as is but the distances-to-camera where computed afterward when `GetDistancesToCamera`.

Alternately, what if the existing depth buffer in the rendering was replaced with the camera distances buffer? The occlusion check would be done with the camera distance, and the depth buffer would be created as needed afterward. As we moved more to ANARI-based rendering, this may be the most efficient path forward.

@jeffamstutz, feel free to weigh in.

DeVasconcelos:
@kmorel Thank you for taking the time to look into the solution and for the feedback!

I agree that it is better to avoid having an additional buffer in the ray tracer. I gave your first suggestion a shot [here in this branch](https://github.com/DeVasconcelos/viskores/tree/transform-depth-buffer-back-to-distance-to-world) where in the `Frame::map()` method that used to call `GetDistancesToCamera`, the original depth buffer is now converted back into distances to the camera in world space through a new worklet. That was the only way I could think of to get these distances without having to do the raytracing all over again in Frame &rarr; Is this what you had in mind?

Personally, I would prefer your second suggestion, where the original depth buffer is replaced with the distances to camera, as there would be no more need to convert (at least in the anari-device). But I don't have an overview of all the code that would need to be adjusted for this change. I can imagine it could get tricky.

jeffamstutz:
I don't see a reason to keep both depths around at the same time -- the ANARI device use of the Viskores rendering system will only ever want world depth, and existing Viskores apps will continue to assume normalized depth. I would just add a switch that writes a normalized value if that is requested (which would default to enabled, as that is status-quo behavior).

viskores/rendering/Canvas.h
```diff
@@ -90,6 +90,14 @@ class VISKORES_RENDERING_EXPORT Canvas
...
   DepthBufferType& GetDepthBuffer();
 
+  /// @brief Get the distances to the camera for each pixel.
+  VISKORES_CONT
+  const DepthBufferType& GetDistancesToCamera() const;
+
+  /// @copydoc GetDistancesToCamera
+  VISKORES_CONT
+  DepthBufferType& GetDistancesToCamera();
+
```
kmorel: I think it would be better to leave `Canvas` unchanged and just add the distances buffer to `CanvasRayTracer`.

kmorel (changes requested):
This looks like the right implementation for the current structure of the code. However, I would prefer not to have the `GetDistancesToCamera` methods added to `Canvas`. This solution will change over time, and I don't want to add something to the interface and then have to worry about having it as a supported part of the interface. Hiding it in `CanvasRayTracer` will make future changes less intrusive to most users.

kmorel:
> I don't see a reason to keep both depths around at the same time -- the ANARI device use of the Viskores rendering system will only ever want world depth, and existing Viskores apps will continue to assume normalized depth. I would just add a switch that writes a normalized value if that is requested (which would default to enabled, as that is status-quo behavior).

I agree, but this is a fallout from the current structure of the code. The ANARI device probably shouldn't be depending on rendering classes outside of `viskores::rendering::raytracing`. The `Canvas` class is one of the few places it does. The best solution would be to move the canvas ray tracing methods into the raytracing module, but that is beyond the scope of this PR. For now, let's put this functionality in `CanvasRayTracer`. Then later we can deprecate that class and move its functionality to `raytracing`.

viskores/rendering/Canvas.h
DeVasconcelos: I have applied the changes you requested and rebased the branch. But I have one doubt with the new approach: Right now,  `Canvas::ResizeBuffers` is simply hidden by `CanvasRayTracer::ResizeBuffers` to make sure that `Canvas` really remains untouched, but the buffers are still all resized by only one method.

Is that okay (since `CanvasRayTracer` will likely be deprecated in the future anyways) or would you prefer me to make `Canvas::ResizeBuffers` virtual and override it properly instead?

kmorel: Yeah, that sounds like a problem.

Go ahead and change `Canvas.h` to make `Canvas::ResizeBuffers` virtual (and mark `CanvasRayTracer::ResizeBuffers` with override).

The problem isn't making changes to `Canvas.h` or that class _per se_. The problem was changing the API. Changing a method to/from virtual won't change the compatibility of code already using the class, so that change is fine.

DeVasconcelos: Alright, done! 

I've also squashed the commits on this branch and rebased with main.

kmorel (approved):
Looks good. Let's get the CI passing and we can merge. Thanks.

PR #423, opened 2026-08-08
dependabot: group updates into a single PR per ecosystem

vicentebolea:
Each dependabot-tracked ecosystem (github-actions, pip in
Utilities/CI, pip in docs/users-guide) currently opens one PR per
outdated dependency. Adds a catch-all group per ecosystem so
dependabot combines them into a single PR instead.


vicentebolea:

PR #427, opened 2026-08-10
CodeQL: exclude conf.py from Python analysis

vicentebolea:
Excludes docs/users-guide/conf.py from CodeQL's Python analysis since it cannot be parsed.


vicentebolea:

PR #428, opened 2026-08-10
Fix CodeQL multiplication-converted-to-larger-type alerts

vicentebolea:
Fixes CodeQL alerts 40-58 ("Multiplication result converted to larger type") by casting operands to the destination width before multiplying.

### DIFF
--- a/viskores/cont/CellLocatorTwoLevel.cxx
+++ b/viskores/cont/CellLocatorTwoLevel.cxx
@@ -333,7 +333,8 @@ struct DimensionsToCount
 {
   VISKORES_EXEC viskores::Id operator()(const DimVec3& dim) const
   {
-    return dim[0] * dim[1] * dim[2];
+    return static_cast<viskores::Id>(dim[0]) * static_cast<viskores::Id>(dim[1]) *
+      static_cast<viskores::Id>(dim[2]);
   }
 };
 
@@ -401,8 +402,9 @@ VISKORES_CONT void CellLocatorTwoLevel::Build()
   binIds.ReleaseResources();
 
   // 6: Compute level-2 dimensions
-  viskores::Id numberOfBins =
-    this->TopLevel.Dimensions[0] * this->TopLevel.Dimensions[1] * this->TopLevel.Dimensions[2];
+  viskores::Id numberOfBins = static_cast<viskores::Id>(this->TopLevel.Dimensions[0]) *
+    static_cast<viskores::Id>(this->TopLevel.Dimensions[1]) *
+    static_cast<viskores::Id>(this->TopLevel.Dimensions[2]);
   viskores::cont::ArrayCopy(viskores::cont::make_ArrayHandleConstant(DimVec3(0), numberOfBins),
                             this->LeafDimensions);
   GenerateBinsL1 generateL1(this->TopLevel.BinSize, this->DensityL2);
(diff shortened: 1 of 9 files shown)


vicentebolea:
@kmorel can you review this?

viskores/cont/CellLocatorTwoLevel.cxx
```diff
@@ -342,7 +342,8 @@ struct DimensionsToCount
 {
   VISKORES_EXEC viskores::Id operator()(const DimVec3& dim) const
   {
-    return dim[0] * dim[1] * dim[2];
+    return static_cast<viskores::Id>(dim[0]) * static_cast<viskores::Id>(dim[1]) *
+      static_cast<viskores::Id>(dim[2]);
```
kmorel, line 337: It probably would have been easier to cast the result than the operands, but this is fine.

kmorel (changes requested):
The updates to loadpng are wrong. That's third party code and should go through the proper respective update mechanism.

I would recommend backing out of those changes for this PR and open a separate PR for those updates.

viskores/cont/CellLocatorTwoLevel.cxx
vicentebolea: The idea is that there can be integer overload during the sum since each elem in dim is 16b by casting them to Id (64b) this goes away. We could probably also just cast the first operand so that it forces casting the rest of them.

vicentebolea:
> The updates to loadpng are wrong. That's third party code and should go through the proper respective update mechanism.
> 
> I would recommend backing out of those changes for this PR and open a separate PR for those updates.

Good catch, find the lodepng changes contained in a patch that I will be submitting to our lodepng fork and upstream

kmorel:
The lodepng changes are now in a patch file and applied by `update.sh`. This sounds better, but still seems wrong. The whole point of having the third-party repo is to manage patches to the software. Having secondary patches applied during the update adds unnecessary complications.

I think the right path forward is to back out of the lodepng changes on this PR completely. We can merge without them. Separately, add your patch to a commit on the [for/viskores branch of the lodepng fork](https://gitlab.kitware.com/third-party/lodepng/-/tree/for/viskores) and update in a separate PR.

viskores/cont/CellLocatorTwoLevel.cxx
kmorel: That makes sense.

vicentebolea:
> The lodepng changes are now in a patch file and applied by `update.sh`. This sounds better, but still seems wrong. The whole point of having the third-party repo is to manage patches to the software. Having secondary patches applied during the update adds unnecessary complications.
> 
> I think the right path forward is to back out of the lodepng changes on this PR completely. We can merge without them. Separately, add your patch to a commit on the [for/viskores branch of the lodepng fork](https://gitlab.kitware.com/third-party/lodepng/-/tree/for/viskores) and update in a separate PR.

MR in third-party repo in process: https://gitlab.kitware.com/third-party/lodepng/-/merge_requests/6

kmorel:
Not for this PR, but I wonder if we should move our third-party forks to GitHub?

vicentebolea:
> Not for this PR, but I wonder if we should move our third-party forks to GitHub?

That would be a good idea for viskores-only dependencies (lodepng is also a VTK dep).

vicentebolea:
@kmorel I decided to bring the lodepng changes later in another PR that can wait for properly bringing these changes upstream/third-parties repo.

kmorel (approved):
Looks good.

PR #432, opened 2026-08-12
Parallelize partitioned global field ranges

meghanto:
This is split from #384 in response to review feedback.

### Summary

- Add an optional threaded local-partition path to `FieldRangeGlobalCompute`.
- Let derived filters reuse the base filter scheduler's device-aware thread count.
- Use the common range implementation from partitioned `Histogram`.
- Initialize `Histogram`'s shared bin width before partition workers start.

### Implementation

The new `FieldRangeGlobalCompute` overload accepts a maximum worker count. Each worker computes ranges from local partitions, the local results are merged after the workers finish, and the existing MPI reduction then produces the global range. The original overload remains sequential and unchanged for existing callers.

`Filter::DetermineNumberOfThreads` is now protected, matching its documented role as an overridable scheduler hook. `Histogram` uses that policy when partition threading is enabled, so OpenMP and TBB execution still avoid nested filter-level threading.

The shared histogram bin width is resolved once before partition workers start. Per-partition executions no longer write it concurrently.

### Testing

- `UnitTestFieldRangeGlobalCompute_mpi` with three ranks and explicit threaded local range computation
- `UnitTestHistogramFilter`
- `UnitTestPartitionedDataSetHistogramFilter`, comparing complete sequential and two-worker histograms and bin widths
- 100 repeated threaded Histogram runs and 20 repeated three-rank `FieldRangeGlobalCompute` runs on a compute node

### DIFF
--- a/viskores/cont/FieldRangeGlobalCompute.cxx
+++ b/viskores/cont/FieldRangeGlobalCompute.cxx
@@ -9,16 +9,45 @@
 #include <viskores/cont/FieldRangeGlobalCompute.h>
 
 #include <viskores/cont/EnvironmentTracker.h>
+#include <viskores/cont/RuntimeDeviceTracker.h>
 
 #include <viskores/thirdparty/diy/diy.h>
 
 #include <algorithm>
+#include <atomic>
 #include <functional>
+#include <future>
+#include <utility>
+#include <vector>
 
 namespace viskores
 {
 namespace cont
 {
+
+namespace
+{
+
+void MergeRanges(std::vector<viskores::Range>& result,
+                 const viskores::cont::ArrayHandle<viskores::Range>& ranges)
+{
+  result.resize(std::max(result.size(), static_cast<std::size_t>(ranges.GetNumberOfValues())));
+  auto portal = ranges.ReadPortal();
+  std::transform(viskores::cont::ArrayPortalToIteratorBegin(portal),
+                 viskores::cont::ArrayPortalToIteratorEnd(portal),
+                 result.begin(),
+                 result.begin(),
+                 std::plus<viskores::Range>());
+}
+
+void MergeRanges(std::vector<viskores::Range>& result, const std::vector<viskores::Range>& ranges)
+{
+  result.resize(std::max(result.size(), ranges.size()));
+  std::transform(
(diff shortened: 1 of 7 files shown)


viskores/cont/FieldRangeGlobalCompute.cxx
```diff
@@ -9,17 +9,71 @@
...
 
 namespace viskores
 {
 namespace cont
 {
 
+namespace
+{
+
+class ScopedThreadFriendlyMemAlloc
```
kmorel: I'm pretty sure you can just use `ScopedRuntimeDeviceTracker` instead of creating a new class.

viskores/cont/FieldRangeGlobalCompute.cxx
```diff
@@ -38,7 +92,57 @@ viskores::cont::ArrayHandle<viskores::Range> FieldRangeGlobalCompute(
...
+  if (threads < 2)
+  {
+    auto lrange = viskores::cont::FieldRangeCompute(pds, name, assoc);
+    return viskores::cont::detail::MergeRangesGlobal(lrange);
+  }
+
+  std::atomic<viskores::Id> nextPartition{ 0 };
+  auto computeRanges = [&]()
+  {
+    ScopedThreadFriendlyMemAlloc threadFriendlyMemAlloc;
```
kmorel: Just use `ScopedRuntimeDeviceTracker`.


```suggestion
    ScopedRuntimeDeviceTracker tracker;
    tracker.SetThreadFriendlyMemAlloc(true);
```

kmorel (changes requested):
Looks pretty good. I suggest a minor change to get rid of the `ScopedThreadFriendlyMemAlloc` class.

Also, looks like you need to [reformat the code](https://github.com/Viskores/viskores/blob/main/CONTRIBUTING.md#reformat-a-topic).

viskores/cont/FieldRangeGlobalCompute.cxx
meghanto: Fixed!

viskores/cont/FieldRangeGlobalCompute.cxx
meghanto: Fixed!

kmorel (approved):
Looks good. We just need to make sure CI passes now. Give me a ping if there is a failure.

PR #433, opened 2026-08-13
Bump sphinx

kmorel:
Bumps the versions of Sphinx packages. This bump was originally requested by @dependabot in PRs like #429. However, a problem with the new version of breathe causes the readthedocs build to timeout (see https://github.com/breathe-doc/breathe/issues/1069). Thus, a patch to breathe is also added.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -17,7 +17,7 @@ build:
   - cmake
   - make
   tools:
-    python: '3.10'
+    python: '3.13'
   jobs:
     post_install:
     - git submodule init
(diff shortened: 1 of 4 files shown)


kmorel:
@vicentebolea Can you review these changes to bump the Python packages for readthedocs? The dependabot kept trying to update them, but the readthedocs build kept failing. It took me hours to track down the problem that is reported at https://github.com/breathe-doc/breathe/issues/1069. This creates a patch that allows the package to be updated.

vicentebolea (approved):
Good catch Ken!

PR #442, opened 2026-08-18
ANARI device: Apply image region and aspect parameters to the perspective camera

DeVasconcelos:
In this PR, I have implemented the handling of the `aspect` and `imageRegion` parameters for the perspective camera of the ANARI device. This is useful, e.g., for a multi-screen setup where setting both parameters is necessary to make sure the transitions between screens are seamless (which is also how we've noticed the parameters are not handled yet in the first place).

**ImageRegion**
The `imageRegion`  parameter is now converted from range [0,1] to [-1,1] through the new `imageRegionToViewport` method and the`PerspectiveRayGen` worklet in `viskores/rendering/raytracing/Camera.cxx` was updated so it can handle viewports which are not full.

**Aspect**
I  turned `m_aspect` into a `std::optional` (as is done in the helide device) and added the new attribute `AspectRatio` to `Camera3DStruct` in `viskores/rendering/raytracing/Camera.h`.
Now, if the user explicitly sets the `aspect` parameter, the ratio is passed directly to the affected worklets/methods in `viskores/rendering/raytracing/Camera.cxx`. Otherwise, `AspectRatio` remains 0 (which is its default to ensure the new changes only affect the ANARI renderer) and the aspect ratio is calculated using the image width and height exactly as before.

As always, I'd be happy to hear your thoughts and any feedback you might have on this approach.

### DIFF
--- a/docs/changelog/anari-apply-imageRegion-and-aspect-parameters.md
+++ b/docs/changelog/anari-apply-imageRegion-and-aspect-parameters.md
@@ -1,0 +1,5 @@
+## Applied imageRegion and aspect parameters to the perspective camera
+
+The ANARI device now applies the `imageRegion` and `aspect` parameters to the perspective camera.
+
+Both parameters are optional: If you do not explictly set `imageRegion` using `anariSetParameter`, the full viewport is rendered by default. If you do not explicitly set `aspect`, the device calculates the aspect ratio automatically using the camera's width and height.
--- a/viskores/rendering/Camera.h
+++ b/viskores/rendering/Camera.h
@@ -346,6 +346,24 @@ public:
   VISKORES_CONT
   viskores::Float32 GetZoom() const { return this->Camera3D.Zoom; }
 
+  /// @brief The aspect ratio of the camera
+  ///
+  /// If the aspect ratio is not explicitly overwritten by `SetAspectRatio`, it will be
+  /// computed from the width and height of the rendered image.
+  ///
+  /// Setting the aspect ratio changes the mode to 3D.
+  ///
+  VISKORES_CONT
+  viskores::Float32 GetAspectRatio() const { return this->Camera3D.AspectRatio; }
+
+  /// @copydoc GetAspectRatio
+  VISKORES_CONT
+  void SetAspectRatio(viskores::Float32 aspect)
+  {
+    this->SetModeTo3D();
+    this->Camera3D.AspectRatio = aspect;
+  }
+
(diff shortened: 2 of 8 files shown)


viskores/rendering/Camera.cxx
```diff
@@ -355,6 +355,14 @@ void Camera::Print() const
...
 viskores::rendering::raytracing::Camera Camera::CreateRaytracingCamera(viskores::Int32 width,
                                                                        viskores::Int32 height) const
+{
+  return this->CreateRaytracingCamera(width, height, this->GetAspectRatio());
+}
+
+viskores::rendering::raytracing::Camera Camera::CreateRaytracingCamera(
+  viskores::Int32 width,
+  viskores::Int32 height,
+  viskores::Float32 aspect) const
```
kmorel: I'm not sure what the point of providing an overloading of this method with an `aspect` argument when there already an `AspectRatio` member.

viskores/rendering/Camera.cxx
```diff
@@ -367,7 +375,9 @@ viskores::rendering::raytracing::Camera Camera::CreateRaytracingCamera(viskores:
   rayCamera.SetHeight(height);
   rayCamera.SetWidth(width);
   rayCamera.SetIsOrthogonalProjection(this->GetMode() == Mode::TwoD);
-  rayCamera.SetCamera3D(this->Camera3D);
+  auto camera3D = this->Camera3D;
+  camera3D.AspectRatio = aspect;
```
kmorel: If you are just setting `this->Camera3D.AspectRatio` to `this->GetAspectRatio()`, there would be no reason to make a copy. In fact, there would be no reason to update the aspect ratio at all.

kmorel (changes requested):
Thanks again for the contribution. I did not run this myself, but I assume it is working correctly.

I just had one comment about an unnecessary argument to set the aspect ratio.

viskores/rendering/Camera.cxx
DeVasconcelos: Ah, you are absolutely right, sorry! This was still a leftover from a  previous approach where I calculated the aspect ratio directly in `CreateRayTracing`, if it was not provided, and since the method was `const`, the copy seemed necessary. I'm afraid, I didn't catch that this has become obsolete during refactoring.

I've just removed the obsolete copy and overload and rebased the branch.

kmorel (approved):
Looks good. Once the CI passes, we can merge this in. If something goes wrong with the CI, alert me in a comment.

Thanks again.

PR #451, opened 2026-09-11
More LaTeX cleanup

kmorel:
This is a couple more instances of fixing old instances of LaTeX commands for the new reStructuredText format that I missed in #450.


kmorel:
@vicentebolea This is a small change covering some LaTeX stuff I didn't get committed last time.

vicentebolea (approved):

PR #453, opened 2026-09-11
Make ray-tracing light position configurable

dpugmire:
This PR removes a hard-coded light offset from the ray tracer that caused its lighting to differ from the ANARI renderer. The light now defaults to the camera position and can be configured through the mapper and scalar-renderer APIs. Existing rendering tests explicitly restore the legacy offset to preserve their baselines, while new tests verify the default and custom light positions.


viskores/rendering/Mapper.cxx
```diff
@@ -143,5 +145,19 @@ void Mapper::SetLogarithmY(bool l)
...
+  this->LightPosition = lightPosition;
+  this->LightPositionSet = true;
+}
+
+viskores::Vec3f_32 Mapper::GetLightPosition() const
+{
+  if (this->LightPositionSet)
+    return this->LightPosition;
+
+  return this->CameraPosition;
```
kmorel, line 160: We should continue to offset the light from the camera. The shading is better when you offset the light.

```suggestion
  viskores::Vec3f_32 scale(2, 2, 2);
  return this->Camera.GetPosition() + scale * this->Camera.GetUp();
```

viskores/rendering/raytracing/RayTracer.cxx
```diff
@@ -220,6 +220,20 @@ Camera& RayTracer::GetCamera()
...
+  this->LightPositionSet = true;
+}
+
+viskores::Vec3f_32 RayTracer::GetLightPosition() const
+{
+  if (this->LightPositionSet)
+    return this->LightPosition;
+
+  return this->camera.GetPosition();
+}
```
kmorel, line 235: Getting the position from the camera should be unnecessary here. Make it a requirement to set the light position (or suffer the consequence of having it stuck at 0, 0, 0). You don't need the `this->LightPositionSet` member.

```suggestion
void RayTracer::SetLightPosition(const viskores::Vec3f_32& lightPosition)
{
  this->LightPositionSet = true;
}

viskores::Vec3f_32 RayTracer::GetLightPosition() const
{
  return this->LightPosition;
}
```

This changes the behavior, but I consider the raytracing code internal, so that should not be a big deal.

docs/changelog/ray-tracer-light-position.md
```diff
@@ -0,0 +1,7 @@
+## Make the ray-tracing light position configurable
+
+The ray tracer and scalar renderer now place the light at the camera by default.
```
kmorel, line 3: As I mentioned before, I think the light should still have the offset. The issue was not the offset. It was that the raytracing was not setting the light properly for the ANARI device.

viskores/rendering/raytracing/ScalarRenderer.cxx
```diff
@@ -229,6 +233,20 @@ void ScalarRenderer::AddField(const viskores::cont::Field& scalarField)
...
+  this->LightPositionSet = true;
+}
+
+viskores::Vec3f_32 ScalarRenderer::GetLightPosition() const
+{
+  if (this->LightPositionSet)
+    return this->LightPosition;
+
+  return this->CurrentCamera.GetPosition();
+}
```
kmorel, line 248: Same comment as before. Don't bother with the camera. Require viskores/rendering/ScalarRenderer to set the light.

```suggestion
void ScalarRenderer::SetLightPosition(const viskores::Vec3f_32& lightPosition)
{
  this->LightPosition = lightPosition;
}

viskores::Vec3f_32 ScalarRenderer::GetLightPosition() const
{
  return this->LightPosition;
}
```

viskores/rendering/raytracing/ScalarRenderer.h
```diff
@@ -32,6 +32,9 @@ class VISKORES_RENDERING_RAYTRACING_EXPORT ScalarRenderer
 protected:
   std::unique_ptr<ShapeIntersector> Intersector;
   std::vector<viskores::cont::Field> Fields;
+  viskores::rendering::raytracing::Camera CurrentCamera;
+  viskores::Vec3f_32 LightPosition;
+  bool LightPositionSet;
```
kmorel, line 37: You don't need `CurrentCamera` nor `LightPositionSet` if you require other classes to set the light position.

```suggestion
  viskores::Vec3f_32 LightPosition;
```

viskores/rendering/testing/UnitTestMultiMapper.cxx
```diff
@@ -47,6 +47,7 @@ void MultiMapperRender(const viskores::cont::DataSet& ds1,
   camera.Azimuth(45.0f);
   camera.Elevation(45.0f);
 
+  mapper1.SetLightPosition(camera.GetPosition() + viskores::Vec3f_32(2, 2, 2) * camera.GetViewUp());
```
kmorel, line 50: You should not need this line. The `viskores::Rendering::Mapper*` classes should set this for you if you don't specify a light position.

viskores/rendering/testlib/RenderTest.cxx
```diff
@@ -204,6 +204,8 @@ void DoRenderTest(viskores::rendering::Canvas& canvas,
       break;
   }
   viskores::rendering::View& view = *viewPointer;
+  view.GetMapper().SetLightPosition(view.GetCamera().GetPosition() +
+                                    viskores::Vec3f_32(2, 2, 2) * view.GetCamera().GetViewUp());
```
kmorel, line 208: Again, this should be unnecessary when the mappers set the light position with the offset.


```suggestion
```

viskores/rendering/Mapper.h
```diff
@@ -64,10 +64,20 @@ class VISKORES_RENDERING_EXPORT Mapper
   virtual void SetLogarithmX(bool l);
   virtual void SetLogarithmY(bool l);
 
+  /// @brief Specify the position of the point light used for shading.
+  ///
+  /// If no position is specified, the current camera position is used.
```
kmorel, line 69:
```suggestion
  /// If no position is specified, the current camera position offset by twice the
  /// view up direction is used.
```

viskores/rendering/ScalarRenderer.cxx
```diff
@@ -59,6 +59,16 @@ void ScalarRenderer::SetDefaultValue(viskores::Float32 value)
...
 }
 
+void ScalarRenderer::SetLightPosition(const viskores::Vec3f_32& lightPosition)
+{
+  this->Internals->Tracer.SetLightPosition(lightPosition);
+}
+
+viskores::Vec3f_32 ScalarRenderer::GetLightPosition() const
+{
+  return this->Internals->Tracer.GetLightPosition();
```
kmorel, line 69: The use of the camera to set the light position should be here instead of `viskores/rendering/raytracing/ScalarRenderer.cxx`.

(That said, I'm not sure what the point of this class is. It does not seem to be used anywhere except testing.)

kmorel (changes requested):
The code in `viskores/rendering/anari-device` that uses `viskores::rendering::raytracing::RayTracer` needs to be updated to set the light.

Currently, the anari-device does not support lights at all. For now, setting the light at the camera should be sufficient.

I note that the anari interop classes are not setting lights at all. Probably the helide device is defaulting to a directional light in the same direction as the camera. Our ray caster does not support directional lights, but a point light at the camera is pretty close.

Eventually, our anari-device should support the `KHR_LIGHT_POINT` and the interop classes should set up at least one light.

PR #458, opened 2026-09-18
Consolidate management of ANARI color arrays

kmorel:
Previously, the handling of color arrays was handled completely separately in the transfer function and image sampler objects of the ANARI device. The change consolidates the implementation of converting every possible type of color supported by ANARI to Viskores colors.

This consolidation also cleans up the color handling for both implementations, making the code more compliant with the specification.


kmorel:
@jeffamstutz Can you review this change to the Viskores ANARI device? I was about to implement the 2D image sample where I realized color array management was about to be duplicated for a third copy. This consolidates the code and introduces some corrections.

cc @dpugmire

kmorel:
@jeffamstutz I lied yesterday. There is one PR that is awaiting your review.

